Issue #7, opened 2024-07-12
Create default Refarch-Apigateway


DanielOber:
Upload the default Refarch-Apigateway that is currently created with Maven-Archetype and remove any internal information related to it.

devtobi:
No longer needed, because Gateway managed in separate repository (https://github.com/it-at-m/refarch).

PR #14, opened 2024-07-17
init apigateway

DanielOber:
**Description**

- added the default Refarch-templated, generated with maven-archetype
- removed the application-*.yml files for dev, kon and prod
- removed gitlab-ci.yml
- changed the intern keycloak-server url with variable
- changed the scm inside pom from gitlab to github

**Reference**

Issues #7


refarch-apigateway/.gitattributes
```diff
@@ -0,0 +1,2 @@
+# Alle Textdateien (* text=auto) sollen den Unix-Zeilenumbruch (eol=lf) nutzen
+* text=auto eol=lf
```
devtobi, line 2: I guess this file was added due to itm-prettier-codeformat. However, it should also affect Java files here. That's not what we want. Maybe we can remove the file completely?

refarch-apigateway/apigateway/.gitignore
devtobi, line 1: Why do we need 2 .gitignore files? I guess it would be better to only use one at top level under refarch-apigateway. The ones in apigateway and frontend should be merged into only 1.

refarch-apigateway/apigateway/src/main/java/refarch/ApiGatewayApplication.java
```diff
@@ -0,0 +1,41 @@
...
+ *      {@link ApiGatewayApplication#main}: e.g.
+ *      <code>System.setProperty("reactor.netty.pool.maxConnections", "1000");</code>.
+ *      </ul>
+ *
+ *      To get more information about Spring Cloud Gateway visit the following link:
+ *
+ * @see <a href=
+ *      "https://cloud.spring.io/spring-cloud-gateway/reference/html/">https://cloud.spring.io/spring-cloud-gateway/reference/html/</a>
+ */
+@SpringBootApplication(scanBasePackages = { "refarch" })
```
devtobi, line 34: I guess the scanBasePackages parameter is obsolete here and should not be needed. Please compare with other projects.

refarch-apigateway/apigateway/src/main/java/refarch/configuration/WebSessionConfiguration.java
```diff
@@ -0,0 +1,117 @@
...
+        joinConfig.getMulticastConfig().setEnabled(false);
+        joinConfig.getTcpIpConfig()
+                .setEnabled(true)
+                .addMember("127.0.0.1");
+
+        return hazelcastConfig;
+    }
+
+    @Bean
+    @Profile({ "dev", "kon", "prod" })
```
devtobi, line 81: References to those profiles should be removed. Maybe we can set this as the default? (And thus for not running locally)?

refarch-apigateway/apigateway/src/main/resources/application-test.yml
```diff
@@ -0,0 +1,63 @@
+wiremock:
+  server:
+    port: 0
+    https-port: -1
+
+spring:
+  
+  # muenchen.de realm is used by default so that the SpringContext can be initalized upon running the tests. Change to your own dev realm if available
+  realm: muenchen.de
```
devtobi, line 9: Having dependency on muenchen.de for running local tests is not acceptable. We need a different solution here (e.g. a local keycloak instance if needed)

refarch-apigateway/apigateway/src/main/resources/application-test.yml
```diff
@@ -0,0 +1,63 @@
...
+            allowedOrigins:
+              - "http://localhost:8081"
+              - "http://127.0.0.1:8081"
+            allowedMethods: "*"
+            allowedHeaders: "*"
+            allowCredentials: true
+            maxAge: 3600
+      routes:
+        - id: sso
+          uri: https://ssotest.muenchen.de/
```
devtobi, line 30: Should be changed to a local instance? Maybe also Mock instance with wiremock is possible?

refarch-apigateway/apigateway/src/main/resources/application-test.yml
```diff
@@ -0,0 +1,63 @@
...
+        - RemoveRequestHeader=x-xsrf-token
+        - TokenRelay=
+
+  # security config
+  security:
+    oauth2:
+      client:
+        provider:
+          keycloak:
+            issuer-uri: https://sso.muenchen.de/auth/realms/${spring.realm}
```
devtobi, line 54: Maybe make sso.muenchen.de a separate parameter and then reuse here?

refarch-apigateway/apigateway/src/main/resources/application-test.yml
```diff
@@ -0,0 +1,63 @@
...
+        registration:
+          keycloak:
+            client-id: refarch
+            client-secret: 0b67133e-f4e0-42f6-b05f-7b51175969e8
```
devtobi, line 58: Is this needed for running tests?

refarch-apigateway/apigateway/src/main/resources/application-test.yml
```diff
@@ -0,0 +1,63 @@
...
+
+  mvc:
+    log-request-details: on
+
+info.appswitcher.url: https://test-url-appswitcher.muenchen.de
```
devtobi, line 63: Is this needed for running tests?

refarch-apigateway/apigateway/src/main/resources/application.yml
```diff
@@ -0,0 +1,58 @@
...
+    info:
+      enabled: on
+    prometheus:
+      enabled: on
+  info:
+    env:
+      enabled: true
+info.application.name: ${spring.application.name}
+info.application.version: @project.version@
+info.appswitcher.url: https://appswitcher.muenchen.de
```
devtobi, line 56: I would remove this reference. Should be set via Helm Charts

refarch-apigateway/frontend/.eslintrc.js
devtobi, line 1: This file must be migrated to flat config format when doing lifecycle for ESLint 9. FlatCompat allows backwards compatibility for incompatible configs. I added an issue for this topic.

refarch-apigateway/frontend/.gitignore
devtobi, line 1: This gitignore file should be removed and combined with the top level .gitignore

refarch-apigateway/frontend/README.md
devtobi, line 1: We should think about where we want to put the template specific documentation and whether we still need READMEs or not. I guess a README file for the refarch-templates project itself would be enough.

refarch-apigateway/frontend/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "eslint-plugin-vuetify": "^2.0.5",
+    "jsdom": "^22.1.0",
+    "prettier": "^3.0.3",
+    "typescript": "^5.2.2",
+    "unplugin-fonts": "^1.0.3",
+    "vite": "^4.5.0",
+    "vite-plugin-vuetify": "^1.0.2",
+    "vitest": "^0.34.6",
+    "vue-tsc": "^1.8.22"
+  }
```
devtobi, line 44: In my opinion we should start pinning the dependency versions. This is especially useful if we start versioning our templates later on. Renovatebot can help managing those dependencies on patch level as well.
Pinning deps will sync it with the behaviour in Java based code. Here dependencies are also pinned.

refarch-apigateway/frontend/pom.xml
```diff
@@ -0,0 +1,151 @@
...
+    <artifactId>refarch-apigateway-frontend</artifactId>
+    <version>0.0.1-SNAPSHOT</version>
+    <packaging>jar</packaging>
+    <name>refarch_apigateway_frontend</name>
+
+
+    <properties>
+        <java.version>11</java.version>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
```
devtobi, line 23: We should update to the latest Java version ASAP

refarch-apigateway/frontend/src/api/LhmAvatarService.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * Der Default-LhmAvatarService, der mucatar.muenchen.de verwendet.
+ *
+ * @see https://git.muenchen.de/km23/mucatar
+ * @see https://github.com/it-at-m/ad2image#documentation
+ */
+export const DefaultLhmAvatarService = new LhmAvatarService(
+  "https://mucatar.muenchen.de/"
+);
```
devtobi, line 31: Maybe we can parameterize this to not be hard coded in here? Env variable should do the trick?

refarch-apigateway/frontend/src/components/common/DatetimeInput.vue
```diff
@@ -0,0 +1,206 @@
...
+ * Beispiel:
+ * <datetime-input
+ *   id="kontakt-field"
+ *   v-model="kontaktDate"
+ *   dense
+ *   clearable
+ *   label="Kontakt am"
+ *   :rules="[(v: string) => !v || moment(v).isBefore(moment.now()) || 'Datum muss in der Vergangenheit liegen.']"
+ * ></datetime-input>
+ */
```
devtobi, line 87: I would not include usage examples in the code.

refarch-apigateway/frontend/src/components/common/YesNoDialog.vue
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * Beispiel:
+ * <yes-no-dialog
+ *    v-model="deleteDialog"
+ *    buttontext="Löschen"
+ *    dialogtitle="Löschen?"
+ *    dialogtext="Wollen Sie die Entität wirklich löschen?"
+ *    @no="deleteDialog = false"
+ *    @yes="deleteSome"></yes-no-dialog>
+ */
```
devtobi, line 79: I would not include usage examples in the code. However we can discuss about this. ;)

refarch-apigateway/frontend/src/composables/formatter.ts
```diff
@@ -0,0 +1,15 @@
+import moment from "moment";
```
devtobi, line 1: We should remove moment as its deprecated. JavaScript API now natively supports the needed features.

refarch-apigateway/frontend/src/plugins/vuetify.ts
```diff
@@ -0,0 +1,20 @@
...
+export default createVuetify({
+  theme: {
+    themes: {
+      light: {
+        colors: {
+          primary: "#333333",
+          secondary: "#FFCC00",
+          accent: "#7BA4D9",
+          success: "#69BE28",
+          error: "#FF0000",
```
devtobi, line 15: Maybe we should make the colors configurable via the initialization tool? Just an idea.. Also, dark mode needs to be supported..

refarch-apigateway/frontend/src/types/User.ts
```diff
@@ -0,0 +1,34 @@
+class User {
+  sub = "";
+
+  // see https://wiki.muenchen.de/betriebshandbuch/wiki/Red_Hat_Single_Sign-On_(Keycloak)#Scopes
```
devtobi, line 4: Remove that link

refarch-apigateway/frontend/src/views/GetStartedView.vue
devtobi, line 1: We should remove all mentioned links or replace them once the documentation has been migrated to Vitepress and made publicly available.

devtobi (changes requested):
Review done! Only added comments to stuff we don't have issues for yet.

refarch-apigateway/.gitattributes
DanielOber: its needed because everyhone has their own git configuration. With that file it is guaranteed that everyone uses `lf` as default line break. Removing it could cause problems.

refarch-apigateway/apigateway/.gitignore
DanielOber: created issue #36

refarch-apigateway/apigateway/src/main/java/refarch/ApiGatewayApplication.java
DanielOber: there are still applications that have the parameter set. I opened a ticket to look into it #37

refarch-apigateway/apigateway/src/main/java/refarch/configuration/WebSessionConfiguration.java
DanielOber: crated issue #38

refarch-apigateway/apigateway/src/main/resources/application-test.yml
DanielOber: maybe we can use [Testcontainers ](https://www.baeldung.com/spring-boot-keycloak-integration-testing). Issues created #39

refarch-apigateway/frontend/package.json
DanielOber: By 'pinning' you mean we should always use the exact version without `^` or `~`? That would mean more LCM-effort for everyone

refarch-apigateway/frontend/pom.xml
DanielOber: created issue #40

refarch-apigateway/frontend/src/components/common/YesNoDialog.vue
DanielOber: created #41. we should discuss this one and maybe move it to the documentation

refarch-apigateway/frontend/src/plugins/vuetify.ts
DanielOber: I think we should actually dont make it configurable, because all application should have a consistent look. If someone wants to change it, then he should do it manually. But yeah dark-mode is missing. For the dark-mode i added #42

refarch-apigateway/frontend/src/composables/formatter.ts
DanielOber: created issue #43

refarch-apigateway/apigateway/src/main/resources/application-test.yml
DanielOber: no, removed

refarch-apigateway/apigateway/src/main/resources/application-test.yml
DanielOber: no, removed

refarch-apigateway/apigateway/src/main/resources/application.yml
DanielOber: added variable for it

refarch-apigateway/apigateway/src/main/resources/application-test.yml
DanielOber: added variable and we like i said in the other comment, maybe testcontainers could help

refarch-apigateway/frontend/README.md
DanielOber: i think using the docs folder would be the best, so everything is in one place.

refarch-apigateway/.gitattributes
devtobi: Makes sense. But shouldnt we add the file to all other templates as well? I will create an issue for it.

devtobi: It might be even better to centralize this file at the repository root. We might be able to do the same with the .gitignore file.

refarch-apigateway/frontend/src/plugins/vuetify.ts
devtobi: Ok! Fine! 😄

refarch-apigateway/frontend/README.md
devtobi: Yes, then we should remove the README file I guess.

refarch-apigateway/frontend/package.json
devtobi: That's true. But not pinning it leads to inconsistent behaviour and might even introduce side effects or problems when a newer version is used without explicitly defining it. We can discuss this at a later point as well. I will create a issue for it.

devtobi:
Package structure should be changed to `de.muenchen.refarch` instead of currently `refarch`.

refarch-apigateway/frontend/README.md
DanielOber: removed

devtobi:
No longer needed, because API-gateway will be managed in https://github.com/it-at-m/refarch. We should only provide a template for the frontend itself.

Issue #56, opened 2024-07-19
Create Default Frontend


DanielOber:
Create a default frontend without the Api-Gateway

PR #61, opened 2024-07-19
init refarch-frontend

DanielOber:
**Description**

- removed api-gateway
- moved .gitattributes into the frontend
- added Mucatar_URL alls enviornment variable

**Reference**

Issues #56


refarch-frontend/vite.config.ts
```diff
@@ -0,0 +1,41 @@
...
+    ViteFonts({
+      google: {
+        families: [
+          {
+            name: "Roboto",
+            styles: "wght@100;300;400;500;700;900",
+          },
+        ],
+      },
+    }),
```
devtobi, line 29: I wonder that we explicitly "import" Roboto font here. Where are the font getting pulled from? Are they included in the final build or loaded externally from Google servers? Some browsers might have them integrated out of the box, but others don't. We shouldn't force loading from external servers and make sure fonts are always included. On top of that, why do we need to define font here? Isn't this Vuetify default?

refarch-frontend/tests/unit/example.spec.ts
devtobi, line 1: Are we going to support component testing in the future? There was not that much consent for unit testing Vue components in the past. We should discuss if we want to further include this.

refarch-frontend/vitest.config.ts
devtobi, line 1: Might not be needed anymore if we remove unit test support. Should be discussed.

refarch-frontend/src/views/HomeView.vue
```diff
@@ -0,0 +1,52 @@
...
+</script>
+
+<style scoped>
+.UP {
+  color: limegreen;
+}
+
+.DOWN {
+  color: lightcoral;
+}
```
devtobi, line 51: Naming CSS classes like this is generally bad practice (see BEM notation). We should consider refactoring this.

refarch-frontend/src/views/GetStartedView.vue
devtobi, line 1: Since here only internal links are included, we should consider removing those and re-adding those once we have useful public documentation.

refarch-frontend/src/types/itm-appswitcher-vue/itm-appswitcher-vue.d.ts
devtobi, line 1: Why is this needed? Isn't the dependency correctly typed? (That's the main purpose for .d.ts files). I suggest trying to remove the file and if it does not work at least collecting type definitions from external libraries in a `declarations.d.ts` file at `src`-root.

refarch-frontend/src/types/User.ts
```diff
@@ -0,0 +1,33 @@
...
+  u.username = "Local Development User";
+  u.displayName = "Local Development User";
+  u.authorities = [
+    // todo add authorities
+  ];
+  u.user_roles = [
+    // todo add user roles
+  ];
+  return u;
+}
```
devtobi, line 31: I think we should refactor this to a static class function instead of a separate function.

refarch-frontend/src/stores/user.ts
```diff
@@ -0,0 +1,21 @@
+import { defineStore } from "pinia";
+import { computed, ref } from "vue";
+
+import User from "@/types/User";
+
+export interface UserState {
+  user: User | null;
```
devtobi, line 7: General topic: `null` handling is kind of irregular in different places and for me union types like `MyType | null` look not very clean. Maybe there is a more TypeScript-native like way to do things?
Sometimes we have `* | null`, sometimes we have `* | undefined`. We should find a consistent solution.

refarch-frontend/src/plugins/vuetify.ts
```diff
@@ -0,0 +1,20 @@
...
+    themes: {
+      light: {
+        colors: {
+          primary: "#333333",
+          secondary: "#FFCC00",
+          accent: "#7BA4D9",
+          success: "#69BE28",
+          error: "#FF0000",
+        },
+      },
```
devtobi, line 17: We could think about separating theme definition into separate file and then import here. Might make the whole theming stuff more clear especially when we add things like dark mode, high contrast mode etc.

refarch-frontend/src/plugins/index.ts
```diff
@@ -0,0 +1,9 @@
+import type { App } from "vue";
+
+import router from "../router";
+import pinia from "./pinia";
+import vuetify from "./vuetify";
```
devtobi, line 5: We should always import files using the `@/..` shorthand provided by Vite. This should also be possible in non Vue SFCs.

refarch-frontend/src/main.ts
```diff
@@ -0,0 +1,10 @@
+import { createApp } from "vue";
+
+import { registerPlugins } from "@/plugins";
+import App from "./App.vue";
```
devtobi, line 4: We should always import files using the `@/..` shorthand provided by Vite. This should also be possible in non Vue SFCs.

refarch-frontend/src/composables/formatter.ts
```diff
@@ -0,0 +1,15 @@
...
+      ? text.toLowerCase().replace(/^\w/, (c) => c.toUpperCase())
+      : "";
+  }
+
+  function formatDate(date: Date): string {
+    return date ? moment(date).format("L") : "";
+  }
+
+  return { startingCharUpperCase, formatDate };
+}
```
devtobi, line 15: This file tries to mimic a Vue composable to expose the mentioned methods. However no Vue specific stuff like Refs etc. is used here so this can be used totally outside the Vue "ecosystem". That's why I suggest to refactor and rename this file (e.g. a static class) to indicate its not Vue dependent.

refarch-frontend/src/components/common/DatetimeInput.vue
devtobi, line 1: Is there no Vuetify native way of combining date with time input?

devtobi (changes requested):
Review done!

refarch-frontend/src/App.vue
```diff
@@ -0,0 +1,139 @@
...
+          cols="6"
+          class="d-flex align-center justify-center"
+        >
+          <v-text-field
+            id="suchfeld"
+            v-model="query"
+            flat
+            variant="solo-inverted"
+            hide-details
+            label="Suche"
```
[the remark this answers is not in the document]
DanielOber: No, the UI itself should still be in german

refarch-frontend/.browserslistrc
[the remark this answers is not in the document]
DanielOber: #66

refarch-frontend/.env.development
```diff
@@ -0,0 +1,2 @@
+VITE_VUE_APP_API_URL=http://localhost:8082
```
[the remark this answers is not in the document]
DanielOber: #67

refarch-frontend/src/App.vue
```diff
@@ -0,0 +1,139 @@
...
+        <v-col
+          cols="3"
+          class="d-flex align-center justify-start"
+        >
+          <v-app-bar-nav-icon @click.stop="drawer = !drawer" />
+          <router-link to="/">
+            <v-toolbar-title class="font-weight-bold">
+              <span class="text-white">RefArch-</span>
+              <span class="text-secondary">Kick</span>
+              <span class="text-white">Starter</span>
```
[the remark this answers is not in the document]
DanielOber: #68

refarch-frontend/src/App.vue
```diff
@@ -0,0 +1,139 @@
+<template>
+  <v-app>
+    <the-snackbar />
+    <v-app-bar color="primary">
+      <v-row align="center">
+        <v-col
+          cols="3"
+          class="d-flex align-center justify-start"
+        >
+          <v-app-bar-nav-icon @click.stop="drawer = !drawer" />
```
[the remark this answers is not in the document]
DanielOber: #69

refarch-frontend/src/App.vue
```diff
@@ -0,0 +1,139 @@
...
+            v-if="appswitcherBaseUrl"
+            :base-url="appswitcherBaseUrl"
+            :tags="['global']"
+          />
+          <v-btn
+            variant="text"
+            icon
+          >
+            <lhm-avatar
+              v-if="userStore.getUser !== null"
```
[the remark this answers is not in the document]
DanielOber: #71

refarch-frontend/src/api/FetchUtils.ts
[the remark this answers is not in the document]
DanielOber: #72

refarch-frontend/src/components/TheSnackbar.vue
```diff
@@ -0,0 +1,63 @@
+<template>
+  <v-snackbar
+    id="snackbar"
+    v-model="show"
+    :color="color"
+    :timeout="timeout"
+  >
+    {{ message }}
+    <v-btn
+      v-if="color === 'error'"
```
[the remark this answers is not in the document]
DanielOber: #74

refarch-frontend/src/views/HomeView.vue
DanielOber: #76

refarch-frontend/vitest.config.ts
DanielOber: #77

refarch-frontend/src/views/GetStartedView.vue
DanielOber: #78

refarch-frontend/src/plugins/vuetify.ts
DanielOber: not sure why we would do this. If you want to change the vuetify theme, you need to know where you have to change it and that should be in the vuetify documentation. Moving thing away from it doesnt sound like a good idea.

refarch-frontend/src/plugins/index.ts
DanielOber: #79

refarch-frontend/src/main.ts
DanielOber: #79

devtobi (approved):
Moved all conversations to separate issues. LGTM now!

Issue #97, opened 2024-07-24
Add CODEOWNERS file and require in PR


devtobi:
**Is your feature request related to a problem? Please describe.**
CODEOWNERS file can be used to automatically assign reviewers to certain files. This is useful if we want to establish code reviews from the core team.

**Describe the solution you'd like**
CODEOWNERS file with wildcard entry (`*`) for all core team members.
Maybe check if we can add a group to CODEOWNERS instead of individual persons.
After the file was added, we can enable the CODEOWNER approval in GitHub rules.

**Describe alternatives you've considered**
Alternatively we could use GitHub Apps like https://probot.github.io/apps/auto-assign/

**Additional context**
---
The CODEOWNERS file can be placed at 3 different locations. Which one should we choose?
1. Repository root
2. `.github` folder
3. `docs` folder

On top of that we need to assign a team as the code owners. Which team should it be?
1. **refarch-admin**
2. **refarch-maintainer**
3. **refarch-devs**
4. All 3 above using **refarch**

devtobi:
I suggest placing the file in the .github folder, to not clutter the main repository
**Note**: All github related files like CODE_OF_CONDUCT, LICENSE, README etc. can be placed inside `.github`. Thats why I suggest moving all files to the new location. GitHub will still correctly find and associate those files.

@DanielOber what do you think about this? Is it ok to move all files in this issue or should we create a separate issue for that?

devtobi:
I activated the branch rule for CODEOWNER approval. :) Will create a follow up issue to refactor GitHub related files into `.github`

PR #112, opened 2024-07-26
:wrench: Added CODEOWNERS file

devtobi:
**Description**
- Added CODEOWNERS file
- Added admin and maintainer group as CODEOWNERS for all files
- Added admin group as CODEOWNERS for CODEOWNERS file itself

**Reference**
Issues #97

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,5 @@
+# These owners will be the default owners for everything in the repo.
+* @it-at-m/refarch-admin @it-at-m/refarch-maintainer
+
+# Protect CODEOWNERS file itself against modification
+/.github/CODEOWNERS @it-at-m/refarch-admin


simonhir (approved):
LGTM but has no effect as long as no branch rules are defined in my opinion

devtobi:
> LGTM but has no effect as long as no branch rules are defined in my opinion

Thats correct, I will activate this after merging. ;)

.github/CODEOWNERS
```diff
@@ -0,0 +1,5 @@
+# These owners will be the default owners for everything in the repo.
+* @it-at-m/refarch-admin @it-at-m/refarch-maintainer
```
simonhir, line 2: @devtobi is there a reason for mentioning admins and maintainers here?

René mentioned he is getting many notifications without really needing his attention, so only maintainers would be better fitting in my opinion.

devtobi: We can definitely change that. I wanted to admins be able to approve as well. Problem is, when we only set maintainers here, admins (even they have more rights than maintainers) will not be able to approve an MR in terms of code ownership.
If admins won't actively participate in development, we can of course remove those from the CODEOWNER file. However the CODE_OWNERS file itself should only be approvable by admins.

simonhir: As far as i know admins are mostly also added to the maintainer group.
I create a MR for that

Issue #120, opened 2024-07-26
Move GitHub related files


devtobi:
**Is your feature request related to a problem? Please describe.**
GitHub related files like LICENSE, README.md, CODE_OF_CONDUCT, ... should be placed inside the `.github` directory to not clutter the repository root.

**Describe the solution you'd like**
Move files to `.github` folder

**Describe alternatives you've considered**
---

**Additional context**
---

PR #121, opened 2024-07-26
:truck: Move GitHub related files

devtobi:
**Description**
Moved GitHub related files to `.github` repository

**Reference**
Issues #120


simonhir (commented):
In my opinion, it is much nicer to leave the files in the root and this is also best practice

DanielOber (approved):
@simonhir 

I had a call with @devtobi and we discussed it again. For now we will move it to the .github folder, because it looks tidier and the files itself wont change often. If there is any confusion later on or People that are using the templates are not happy with it, we will change it to the Top Level again.

DanielOber:
> LICENSE isn't detected by GitHub anymore

yeah you were right. Issue https://github.com/it-at-m/refarch-templates/issues/123
MR for the fix: https://github.com/it-at-m/refarch-templates/pull/124

Issue #11, opened 2024-07-12
Add issue-Templates


DanielOber:
- what templates do we need
- do we need to change existing templates

devtobi:
We discussed this today.
We want to include at least issue templates for:
- Bugs (Type: Bug, Priority: High)
- Docs (Type: Documentation, Priority: Low)
- Feature Request (Type: Enhancement, Priority: Medium)
- Maintenance Request (Type: Maintenance: Priority: Medium)
- Question/Support (Type: Question, Priority: High)

@DanielOber Do you agree with this?

https://github.com/Josee9988/project-template/tree/master/.github/ISSUE_TEMPLATE has some really nice template defaults we can use.

DanielOber:
Yes 👍

devtobi:
There were some small "errors" in the template. Created a follow-up bug-issue under https://github.com/it-at-m/refarch-templates/issues/131

PR #126, opened 2024-07-26
Docs/11 add issue templates

devtobi:
**Description**
- Added issue template "Bug Report"
- Added issue template "Feature Request"
- Added issue template "Maintenance work"
- Added issue template "Documentation change"
- Added template chooser configuration

**Reference**
Issues #11


simonhir (approved):
LGTM but not really sure if the "maintenance" templates is a real use case and if not all "maintenance" cases could be done as feature request  or bug.

DanielOber:
> LGTM but not really sure if the "maintenance" templates is a real use case and if not all "maintenance" cases could be done as feature request or bug.

we wanted to base it on the labels we have, so there are issues templates for maintanance and feature separately

Issue #10, opened 2024-07-12
Create Docker-Compose


DanielOber:
Create a new directory within the refarch-template repository named stack, and include a docker-compose.yml file inside it. This file should contain postgres, pgadmin, keycloak, keycloakmigration and API-Gateway. You can refer to the [digiwf ](https://github.com/it-at-m/digiwf-core/tree/dev/stack)project on GitHub for guidance on how to write the file and what else ist needed for the keycloackmigration.

The local profile must be reconfigured to use the running docker services.

devtobi:
Should this really be at the top level of the repo? Or is it something that's specific to the backend?

DanielOber:
i think it should be top level, because it could be needed by every service and not only by a specific specific. For example, if you add a s3 bucket (minio) then the EAI / Frontend need to access it.

PR #134, opened 2024-07-30
feat/10-create-docker-compose

DanielOber:
**Description**

- added docker compose for local development
- used [Wahllokalsystem](https://github.com/it-at-m/Wahllokalsystem/tree/dev/stack) and [DigiWf](https://github.com/it-at-m/digiwf-core/tree/dev/stack) as reference
**Reference**
- added keycloak, keycloak-init, keycloak-postgres, postgres, pgadmin and the new refarch-gateway as default services. 

Issues #10


stack/docker-compose.yml
```diff
@@ -0,0 +1,94 @@
+name: default-services
+
+services:
+  postgres:
+    container_name: postgres
+    image: postgres:16.0-alpine3.18
```
devtobi, line 6: Since Docker Compose 3.4 you can create reusable variables like this:

```
x-versions:
  postgres: &postgres_version 13.4
  redis: &redis_version 6.2.5
  app: &app_version 1.0.0
```

This would allow us to centralize versions of images.

To reuse variables you can use e.g:
```
image: postgres: *postgres_version
```

stack/docker-compose.yml
```diff
@@ -0,0 +1,94 @@
...
+    environment:
+      - POSTGRES_DB=postgres
+      - POSTGRES_USER=admin
+      - POSTGRES_PASSWORD=admin
```
devtobi, line 10: Why are some environment variables injected via .env file and some hardcoded? We should find a standardized way.

stack/docker-compose.yml
```diff
@@ -0,0 +1,94 @@
...
+    ports:
+      - "5432:5432"
+
+  pg-admin:
+    container_name: pg-admin
+    image: dpage/pgadmin4
```
devtobi, line 16: Please include a concrete image tag as otherwise always latest is used. In my opinion latest is bad practice because its not deterministic.
Tools like renovatebot will easily allow us to managed image version bumps.

stack/docker-compose.yml
```diff
@@ -0,0 +1,94 @@
...
+    environment:
+      - PGADMIN_DEFAULT_EMAIL=example@example.com
+      - PGADMIN_DEFAULT_PASSWORD=admin
```
devtobi, line 19: I suggest moving all env variables into .env files. If the env variables are only needed by a particular service we can create e.g. db.env, etc.

stack/docker-compose.yml
```diff
@@ -0,0 +1,94 @@
...
+    image: quay.io/keycloak/keycloak:20.0.3
+    command:
+      - start-dev --http-relative-path /auth
+    depends_on:
+      - db-postgres-keycloak
+    environment:
+      - KEYCLOAK_ADMIN=${KEYCLOAK_ADMIN}
+      - KEYCLOAK_ADMIN_PASSWORD=${KEYCLOAK_ADMIN_PASSWORD}
+      - KC_HOSTNAME_STRICT=false
+      - DB=postgres
```
devtobi, line 34: This value should be directly linked to the value set in the db service. Otherwise manual adjustments in multiple service definitions are necessary.

stack/docker-compose.yml
```diff
@@ -0,0 +1,94 @@
...
+      WAIT_FOR_KEYCLOAK: 'true'
+      KEYCLOAK_CHANGELOG: /migration/keycloak-changelog.yml
+    volumes:
+      - './keycloak/migration:/migration'
+    networks:
+      - keycloak
+
+  refarch-gateway:
+    container_name: refarch-gateway
+    image: ghcr.io/it-at-m/refarch/refarch-gateway:latest
```
devtobi, line 73: I highly discourage the use of latest image.
We as template providers should make sure our templates work with a specific version if the API gateway. If we set latest here, bumping the API-Gateway of the stack is no longer under our control. This might lead to side effects.
I suggest we manually bump API-Gateway versions as its a external dependency just like db, mail, etc.

devtobi (changes requested):

stack/docker-compose.yml
DanielOber: one postgres ist explicit for the keycloak, the other is for local development. So if we bump the postgres version, both will get updated. If the keycloak version is not compatible with the postgres version there will be problems. And both postgresdbs have different use cases so either way i dont want to have any coupling between those two.

stack/docker-compose.yml
DanielOber: it depends if the variables have to be reused or not. 
Example:
The Postgres_DB variable you linked here is not used anywhere else
The SSO_REALM on the other hand is used in several places

stack/docker-compose.yml
DanielOber: dont know if an env file with 30 differen variables will help anyone or will make anything better

stack/docker-compose.yml
DanielOber: not exactly sure you want to do that. do you have any example

stack/docker-compose.yml
devtobi: Yeah that's fine. However I still suggest defining the image versions used at the top level of the docker-compose, for better maintainability.
We have the same concept in our pom.xml files

stack/docker-compose.yml
devtobi: I would encourage it because it decouples the static docker configuration from runtime relevant environment variables. This way if we want to change runtime stuff, we directly know we can look it up in the env files.

stack/.env
```diff
@@ -0,0 +1,10 @@
+SSO_BASE_URL=http://keycloak:8080/auth
```
simonhir, line 1: Did you think about alternatives for `keycloak`?
Think it might be a problem as with this every developer need at least one time local administration rights.

DanielOber: >  local administration rights

why?

simonhir: `keycloak` needs to be added to hosts file that the login redirect works.

stack/docker-compose.yml
devtobi: Renovate is currently not capable of scanning the versions in a custom defined block (see https://github.com/nodeca/js-yaml?tab=readme-ov-file#caveats). Thats why need to drop this temporary.

stack/.env
DanielOber: yeah okay, i understand what you mean. I will look into it. Changing the hosts file just to make it work is rather bad

devtobi:
We discussed this and we are going to remove `.env` file for now (as it can only be used easily inside IntelliJ) and handle everything inside `docker-compose.yml` itself. Custom Anchors for reusability inside the `docker-compose.yml` are optional.

devtobi:
@DanielOber Should we do another local test of the docker compose stack before merging?

DanielOber:
> @DanielOber Should we do another local test of the docker compose stack before merging?

yes please

stack/docker-compose.yml
devtobi, line 1: Please add `restart: unless-stopped` to each container. Otherwise the fail of a healthcheck does not trigger a restart of the container.

stack/docker-compose.yml
```diff
@@ -0,0 +1,120 @@
...
+      - KEYCLOAK_ADMIN=admin
+      - KEYCLOAK_ADMIN_PASSWORD=admin
+      - KC_HOSTNAME_STRICT=false
+      - KC_HEALTH_ENABLED=true
+      - DB=postgres
+      - DB_URL=jdbc:postgresql://db-postgres-keycloak:5432/keycloak
+      - DB_USERNAME=keycloak-user
+      - DB_PASSWORD=keycloak-secret
+    ports:
+      - "8080:8080"
```
devtobi, line 53: Personally I don't think its good to block port 8080 locally. We should use `8100`. Might be enough to change the host port only e.g. "8100:8080".

stack/keycloak/migration/03_add-user.yml
devtobi, line 1: I think its easier to remember if the user is just named `user` (user@user.com) with password `user`. Then it is easier to remember. Firstname and Lastname can both be `user`

devtobi (changes requested):

devtobi (changes requested):
Oh and please take a look at
grafik
I get those issues inside the console for both database containers and they don't look very healthy. ;)

stack/docker-compose.yml
DanielOber: `unless-stopped` sounds like a bad idea, would prefer to go with `on-failure[:max-retries]`. Other wise it could end in an endless loop. It the same settings we use in openshift

devtobi: Yes that sounds like a good idea. :)

DanielOber:
> Oh and please take a look at grafik I get those issues inside the console for both database containers and they don't look very healthy. ;)

the health check didnt use the user and db name, thats why the error came. Wasnt really a problem, but i fixed it. 

pls try again, should work now

devtobi (approved):
LGTM now :)

Issue #122, opened 2024-07-26
Remove eslint-plugin-vuetify


devtobi:
**Is your feature request related to a problem? Please describe.**
`eslint-plugin-vuetify` is only used for migration from Vuetify 2 to 3. 

**Describe the solution you'd like**
Remove `eslint-plugin-vuetify` from dependencies. Also remove references in ESLint configuration

**Describe alternatives you've considered**
---

**Additional context**
---

PR #137, opened 2024-07-31
Maintenance/122 remove eslint plugin vuetify

devtobi:
**Description**
- Removed eslint-plugin-vuetify because its purpose as a migration tool is no longer needed
- Updated lock-file (no `resolved`-key anymore) -> correct local `.npmrc` configuration needed (`omit-lockfile-registry-resolved=true`)
- Reformatted frontend files with new eslintconfig
- Added custom `.npmrc` to local project setup to make sure `resolved` property in lockfile is omitted

**Reference**
Issues #122


DanielOber (approved):
LGTM

Issue #139, opened 2024-07-31
[Maintenance Work] Add PR labeler workflow


devtobi:
### Relevant template

None

### Desired solution

To better categorize the created PRs a GitHub action should be used to automatically label the PRs depending on branch name, changed files etc.
A GitHub Workflow is needed as well as a configuration file for the labeling strategy

### Additional context (optional)

See https://github.com/it-at-m/refarch/pull/43/files as reference.

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #143, opened 2024-08-01
Maintenance/139 maintenance work add pr labeler workflow

devtobi:
**Description**
- Added PR labeler workflow, keeps labels in sync when head branch is updated
- Added PR labeler configuration using Type and Template Labels
   - supports branch prefixes
   - supports automatic branch names generated via issue title (based on the available issue forms)
   - supports checking of changed files for template labels

**Reference**
Issues #139

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,26 @@
+"Type: Feature":
+  - head-branch: [ '^feat', 'feature' ]
+"Type: Bug":
+  - head-branch: [ '^bug', 'bug' ]
+"Type: Maintenance":
+  - head-branch: [ '^maint', 'maintenance' ]
+"Type: Documentation":
+  - head-branch: [ '^docs', "documentation" ]
+  - changed-files:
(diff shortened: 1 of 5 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,26 @@
+"Type: Feature":
+  - head-branch: [ '^feat', 'feature' ]
+"Type: Bug":
+  - head-branch: [ '^bug', 'bug' ]
+"Type: Maintenance":
+  - head-branch: [ '^maint', 'maintenance' ]
```
simonhir, line 6:
```suggestion
  - head-branch: [ '^maint', 'maintenance', '^refactor' ]
```

Don't know about you but i often use the refactor prefix

simonhir (approved):

.github/labeler.yml
devtobi: We want the branch names to be oriented on the `Type` labels. That's why we explicitely keep `refactor` out. But we discussed this. :D

Issue #140, opened 2024-07-31
[Maintenance Work] Update version properties naming in Maven based projects


devtobi:
### Relevant template

refarch-backend

### Problem description (optional)

Renovatebot is capable of updating Version numbers defined in a custom maven property.
However this only works when the property ends with `.version`. (see https://docs.renovatebot.com/presets-customManagers/#custommanagersmavenpropertyversions).
Currently we have a mix.

### Desired solution

Refactor all properties for dependency versions to end on `.version`

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #144, opened 2024-08-01
Maint/140 maintenance work update version properties naming in maven based projects

devtobi:
**Description**
- Updated custom maven properties for dependency versions to end in `.version`
- Updated naming scheme of custom properties to `artefactName.version`
- Sorted dependencies by use case

**Reference**
Issues #140


simonhir (approved):
LGTM

Issue #20, opened 2024-07-17
Java Code Linting


devtobi:
**Is your feature request related to a problem? Please describe.**
Currently only Spotless is included to standardize formatting of Java-based code. However a linter should be included in the build process as well. This helps which finding code smells and bugs.

**Describe the solution you'd like**
A solution would be a tool like [PMD](https://docs.pmd-code.org/latest/index.html) or [SpotBugs](https://spotbugs.readthedocs.io/en/latest/index.html). Both are available as a Maven plugin.

When a solution is chosen, IDE Plugins should be supported as well.

**Describe alternatives you've considered**
---

**Additional context**
---

DanielOber:
We got SonarLint, why do we need another solution?

devtobi:
Is SonarLint integrated in the build chain? No. The usage of SonarLint depends on an external service (or an IDE plugin) and thus will not work under all circumstances. That's why I suggested to include PMD or SpotBugs that's directly integrated into the Maven build process.
Additionally, not all projects are using SonarLint responsively. That's another reason why we should enforce the use of such tools and make those available locally as well :)

devtobi:
Yeah great, that's good. However I don't see it as a "one or the other" solution. For frontend projects we have ESLint in place. This a tool for code linting. It's embedded inside our build process. You could argue we don't need ESLint because SonarLint can do same things. Linting inside CI/CD and locally is not the same. That's why I suggest to add PMD or SpotBugs, independently of which solution we come up with for CI/CD.

devtobi:
Follow up will be done in: https://github.com/it-at-m/refarch-templates/issues/177 due to a configuration problem.

PR #151, opened 2024-08-01
feat/20-java-code-linting

DanielOber:
**Description**

- added spotbugs and pmd

To Try it, first run mvn compile and afterwards run mvn site in the backend. In the target folder it will create an index.html that you can open. Afterwards you will see the gui with the folder on the left site, that says "projektbericht" (its german currently). There you can see the resuls from pmd and spotbugs. Check it out and tell me what you think about the results. 

**Reference**

Issues #20


devtobi:
I took a look at the PMD configuration and would suggest the following changes:
- Remove aggregate reports for CPD and PMD as we don't execute it in the scope of a maven multi-module project (see https://maven.apache.org/plugins/maven-pmd-plugin/examples/removeReport.html) to disable all aggregate-reports
- Add `multithreading.xml` and `performance.xml` to the rulesets used (see https://docs.pmd-code.org/pmd-doc-7.3.0/pmd_rules_java.html)
- Make the build fail when PMD has errors by inlcuding in the check phase -> See https://maven.apache.org/plugins/maven-pmd-plugin/examples/violationChecking.html
- Maybe maintain a "ignore" file for specific rules that are unrealistic in our scenario? -> https://maven.apache.org/plugins/maven-pmd-plugin/examples/violation-exclusions.html
- Maybe a less strict configuration for tests?: https://maven.apache.org/plugins/maven-pmd-plugin/examples/differentRulesetForTests.html, however we could start with the same config for the whole codebase and change later on.

devtobi:
I top a look at the SpotBugs configuration and would suggest the following changes:
- Explicitely set the effort configuration for the SpotBugs analysis (see https://spotbugs.readthedocs.io/en/stable/effort.html), settings this to `Max` should not be an issue as time is not a factor
- Omit certain bug detectors if applicable using https://spotbugs.github.io/spotbugs-maven-plugin/usage.html#specifying-which-bug-detectors-to-skip
- Use additional detectors via https://spotbugs.github.io/spotbugs-maven-plugin/usage.html#using-third-party-or-your-own-detectors, useful additional detectors recommended by SpotBugs listed under https://spotbugs.github.io/#extensions (including analysis of OWASP Top 10, ...)
- Enable failing of the build by using in `check` goal (see https://spotbugs.github.io/spotbugs-maven-plugin/examples/violationChecking.html)

As far as the reported violations I guess we need to check if we can fix them and maybe add some ignore rules later on.

On top of that please refactor the version of SpotBugs used into a maven property `spotbugs-maven-plugin.version`

devtobi (changes requested):
See changes in comments

DanielOber:
> I took a look at the PMD configuration and would suggest the following changes:
> 
>     * Remove aggregate reports for CPD and PMD as we don't execute it in the scope of a maven multi-module project (see https://maven.apache.org/plugins/maven-pmd-plugin/examples/removeReport.html) to disable all aggregate-reports
> 
>     * Add `multithreading.xml` and `performance.xml` to the rulesets used (see https://docs.pmd-code.org/pmd-doc-7.3.0/pmd_rules_java.html)
> 
>     * Make the build fail when PMD has errors by inlcuding in the check phase -> See https://maven.apache.org/plugins/maven-pmd-plugin/examples/violationChecking.html
> 
>     * Maybe maintain a "ignore" file for specific rules that are unrealistic in our scenario? -> https://maven.apache.org/plugins/maven-pmd-plugin/examples/violation-exclusions.html
> 
>     * Maybe a less strict configuration for tests?: https://maven.apache.org/plugins/maven-pmd-plugin/examples/differentRulesetForTests.html, however we could start with the same config for the whole codebase and change later on.
> 
> 
> As far as the reported violations, I really like it. We could maybe disable stuff like `CommentRequired`, `CommentSize` ( or maybe the whole `documentation.xml`?), `LongVariable`
> 
> On top of that please refactor the version of PMD used into a maven property `maven-pmd-plugin.version`

1. yeah goot point, i will remove cpd
2. i would add `performance.xml` but not `multithreading.xml`. Nearly none of our project use multithreading, so i would not add it as default
3. i actually think, it should be the same as in eslint. It should help to improve code but not make it necessary to add every suggestion. If we do it, people will just remove the tool.
4. & 5. like you said, maybe just go with the config for now and make the changes if we see the need for it.

wanna just try it with the documentation and we also change it needed or wanna remove it now and add it later if we need it.

devtobi:
> > I took a look at the PMD configuration and would suggest the following changes:
> > ```
> > * Remove aggregate reports for CPD and PMD as we don't execute it in the scope of a maven multi-module project (see https://maven.apache.org/plugins/maven-pmd-plugin/examples/removeReport.html) to disable all aggregate-reports
> > 
> > * Add `multithreading.xml` and `performance.xml` to the rulesets used (see https://docs.pmd-code.org/pmd-doc-7.3.0/pmd_rules_java.html)
> > 
> > * Make the build fail when PMD has errors by inlcuding in the check phase -> See https://maven.apache.org/plugins/maven-pmd-plugin/examples/violationChecking.html
> > 
> > * Maybe maintain a "ignore" file for specific rules that are unrealistic in our scenario? -> https://maven.apache.org/plugins/maven-pmd-plugin/examples/violation-exclusions.html
> > 
> > * Maybe a less strict configuration for tests?: https://maven.apache.org/plugins/maven-pmd-plugin/examples/differentRulesetForTests.html, however we could start with the same config for the whole codebase and change later on.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > As far as the reported violations, I really like it. We could maybe disable stuff like `CommentRequired`, `CommentSize` ( or maybe the whole `documentation.xml`?), `LongVariable`
> > On top of that please refactor the version of PMD used into a maven property `maven-pmd-plugin.version`
> 
>     1. yeah goot point, i will remove cpd
> 
>     2. i would add `performance.xml` but not `multithreading.xml`. Nearly none of our project use multithreading, so i would not add it as default
> 
>     3. i actually think, it should be the same as in eslint. It should help to improve code but not make it necessary to add every suggestion. If we do it, people will just remove the tool.
> 
>     4. & 5. like you said, maybe just go with the config for now and make the changes if we see the need for it.
> 
> 
> wanna just try it with the documentation and we also change it needed or wanna remove it now and add it later if we need it.

1. Please dont Remove CPD itself (as it is a useful feature), just the aggregate reports for both CPD and PMD.
2. If we don't have multithreaded code providing the multithreading rules shouldn't bother the code. And when users decide to add multi-threaded code they don't have to do another setup. Thats why I would still add it. It does not hurt.
3. Difference is for frontend code with ESLint the CI/CD pipeline fails. I have the strong assumption that no one will ever take a look into the generated reports. That's why I suggest to include it directly in `<build>` instead of `<reporting>`.
We could however start by only doing reporting and making sure the pipeline still fails.
Long term I think it would be better to let the build fail in both JavaScript and Java-based projects. This would even allow us to remove the direct call of linting and formatting tools inside CI/CD because they are automatically invoked when building.

DanielOber:
> Omit certain bug detectors if applicable using 

with that i would actually wait, till we fix the reported errors. Not sure which one we should ignore. 

> Use additional detectors vi

dont think we should everything. I added the findsecbug plugin. The other one is not that popular and it looks like that there is not much happening in the repository anymore. over 60 open issues and last commit 4 months ago. I think its fine if we just use one.

DanielOber:
> Please dont Remove CPD itself (as it is a useful feature), just the aggregate reports for both CPD and PMD. 

i changed it from generating a report to directily showing it in the CL. I think that should be enough, so you can just run the goal you want 

> f we don't have multithreaded code providing the multithreading rules shouldn't bother the code. And when users decide to add multi-threaded code they don't have to do another setup. Thats why I would still add it. It does not hurt.

yeah okay fine by me. 

> pipeline still fails

the point is, that not everything that pmd reports needs to be fixed, thats also why we added some exceptions. Bigger projects will run into more other/more pmd errors that will maybe not make sense. That means, they have to change the pmd-rulesets by themselves. If you have to configure these things by your self all the time, you will get annoyed. Thats why i would prefer setting it to warning and not to error. 

I added it to the it now, that the verify goal fails, if pmd or spotbugs errors exist. But i think we have to talk about that again.

devtobi (approved):

Issue #43, opened 2024-07-17
Remove moment.js


DanielOber:
**Is your feature request related to a problem? Please describe.**
Moment.js is deprecated and even on their site they suggest do move to something else. 

**Describe the solution you'd like**
Remove moment.js and use the js native alternatives

PR #163, opened 2024-08-07
Maint/43 remove momentjs

devtobi:
**Description**
- Refactored formatter file to export individual function (it was never a Vue composable)
- Refactored formatDate function to not rely on moment.js
- Removed moment.js from dependencies

**Reference**
Issues #43


refarch-frontend/src/util/formatter.ts
```diff
@@ -0,0 +1,15 @@
+export function toFirstLetterUppercase(text: string): string {
+  return text ? text[0].toUpperCase() + text.slice(1).toLowerCase() : "";
```
simonhir, line 2: Could fail if only one char long.

simonhir (approved):
LGTM

PR #164, opened 2024-08-07
Maint/refactor function

devtobi:
**Description**
- Refactored occurrences of arrow functions to use function keyword instead

**Reference**
None


simonhir (approved):
LGTM

Issue #117, opened 2024-07-26
Move entrypoint for Vue router


devtobi:
**Is your feature request related to a problem? Please describe.**
Vue router is currently defined in `index.ts` inside `router` directory.
This file cannot be easily search for.

**Describe the solution you'd like**
Router entrypoint should be placed in `plugins` directory with the name `router.ts`. This allows for easier finding.

**Describe alternatives you've considered**
---

**Additional context**
---

PR #165, opened 2024-08-07
:recycle: Moved router to plugins

devtobi:
**Description**
- Moved entrypoint for Vue router to `plugins` folder under `router.ts`

**Reference**
Issues #117


simonhir (approved):
Don't like the reasoning with easier finding but seems to be the better place anyway.
LGTM

PR #167, opened 2024-08-07
Feature/update pom parent

### LINKED ISSUE #19: Remove muenchen-pom
**Is your feature request related to a problem? Please describe.**
At the moment, the muenchen-pom used as parent inside our poms is internal. There is no real reason why we need it and it was last maintained 3 years ago.

**Describe the solution you'd like**
remove the muenchen-pom

simonhir:
**Description**

- Use spring-boot-starter-parent as parent for backend and eai
- Remove build logic already existing in parent
- Fix dependency versions and vars

**Resources**

Issue closes #19


refarch-backend/pom.xml
devtobi, line 1: Following additional points:
- I guess we dont need `maven-compiler-plugin.version` as its included in the parent
- I guess we don't need `maven-resources-plugin.version` as its included in the parent
- We should update the comment next to `lombok.version` to point to spring-boot version defined in parent
- We should add the license and developers sections as you suggested, should we do that in a separate issue?

refarch-eai/pom.xml
```diff
@@ -20,13 +21,11 @@
         <maven.compiler.source>${java.version}</maven.compiler.source>
         <maven.compiler.target>${java.version}</maven.compiler.target>
         <maven.compiler.release>${java.version}</maven.compiler.release>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven-scm-plugin.version>1.10.0</maven-scm-plugin.version>
         <maven-compiler-plugin.version>3.8.1</maven-compiler-plugin.version>
 
         <!-- Core Frameworks -->
         <camel-spring-boot-dependencies.version>4.1.0</camel-spring-boot-dependencies.version>
```
devtobi, line 28: We should add a note stating this must match the spring-boot version provided in the parent.

refarch-eai/pom.xml
```diff
@@ -20,13 +21,11 @@
         <maven.compiler.source>${java.version}</maven.compiler.source>
         <maven.compiler.target>${java.version}</maven.compiler.target>
         <maven.compiler.release>${java.version}</maven.compiler.release>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
         <maven-scm-plugin.version>1.10.0</maven-scm-plugin.version>
         <maven-compiler-plugin.version>3.8.1</maven-compiler-plugin.version>
```
devtobi, line 25: Maybe we can remove both those properties?

refarch-eai/pom.xml
devtobi, line 1: Following additional points:
- Do we need maven-release-plugin? Its not used in the backend as far as I've seen, so why need it in EAI?
- Maybe remove maven-scm-plugin, as far as I've seen it only provides goals to fetch VCS data. We don't need this functionality

devtobi (changes requested):

refarch-eai/pom.xml
simonhir: Removed both

refarch-eai/pom.xml
simonhir: Removed both

refarch-backend/pom.xml
simonhir: All done.
Additional information i.e. for license are currently not needed.

devtobi (approved):

Issue #118, opened 2024-07-26
Jump to top when navigating with Vue router


devtobi:
**Is your feature request related to a problem? Please describe.**
At the moment the Vue router remembers the scroll positions when navigating the app.
It would be better if it would always jump to the top and left when routing to a new page.

**Describe the solution you'd like**
See https://router.vuejs.org/guide/advanced/scroll-behavior.

**Describe alternatives you've considered**
---

**Additional context**
---

PR #168, opened 2024-08-07
:sparkles: Added jumping to top when routing

devtobi:
**Description**
- Added router scroll behaviour to instantly jump back to top and left when navigating pages

**Reference**
Issues #118


refarch-frontend/src/plugins/router.ts
simonhir, line 1: I think to return to the last position on return instead of top seems more natural and is default behavior.

DanielOber:
Lets try it for now, so that everyone atleast know it exits. Later on we can also put in the the documentation so that there is no confusion.

DanielOber (approved):

Issue #153, opened 2024-08-02
[Maintenance] Adjust local profile to match local Docker stack


devtobi:
### Relevant template

refarch-backend

### Problem description (optional)

Now that we have a local Docker stack up and running to start up required development services we need to adjust the local configuration of the backend to match the local Docker setup.

### Desired solution

Adjust local profile to correct DB and Keycloak connection settings

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #170, opened 2024-08-08
Feature/backend local conf

simonhir:
**Description**

Setup backend for local use with stack.

**Reference**

Issues #153


refarch-backend/pom.xml
```diff
@@ -134,14 +134,14 @@
...
 
         <!-- Database -->
-        <dependency>
-            <groupId>com.h2database</groupId>
-            <artifactId>h2</artifactId>
-        </dependency>
         <dependency>
             <groupId>com.mysql</groupId>
             <artifactId>mysql-connector-j</artifactId>
         </dependency>
```
devtobi, line 140: Is this JDBC driver also used for Postgres?

I checked the Postgres homepage and they suggest using only `org.postgresql.postgreql` which you already added. I guess we can remove this.

refarch-backend/src/main/resources/application-test.yml
devtobi, line 1: As far as I've seen you removed this file because all properties are also set in the `local` profile. Is that right?

refarch-backend/src/main/resources/application.yml
```diff
@@ -7,8 +7,17 @@ spring:
         jwt:
           audiences:
           # adds audience verification - https://docs.spring.io/spring-security/reference/servlet/oauth2/resource-server/jwt.html#_supplying_audiences
+          # requires client to be in aud claim, see https://www.keycloak.org/docs/latest/server_admin/#_audience_resolve
```
devtobi, line 10:
```suggestion
          # requires client to be in audience claim, see https://www.keycloak.org/docs/latest/server_admin/#_audience_resolve
```

stack/docker-compose.yml
```diff
@@ -107,14 +107,16 @@ services:
     - SSO_CLIENT_ID=local
     - SSO_CLIENT_SECRET=client_secret
     - SPRING_CLOUD_GATEWAY_ROUTES_0_ID=backend
-    - SPRING_CLOUD_GATEWAY_ROUTES_0_URI=http://backend-service:39146/
+    - SPRING_CLOUD_GATEWAY_ROUTES_0_URI=http://host.docker.internal:39146/
```
devtobi, line 110: Why is this needed? Can't the API-Gateway talk to the backend because there are not in the same Docker network and thus, this special DNS name by docker is necessary?

stack/docker-compose.yml
```diff
@@ -107,14 +107,16 @@ services:
...
     - SPRING_CLOUD_GATEWAY_ROUTES_0_PREDICATES_0=Path=/api/backend-service/**
     - SPRING_CLOUD_GATEWAY_ROUTES_0_FILTERS_0=RewritePath=/api/backend-service/(?<urlsegments>.*), /$\{urlsegments}
     - ALLOWED_ORIGINS_PUBLIC=http://localhost:*
     - ALLOWED_ORIGINS_CLIENTS=http://localhost:*
     - SPRING_PROFILES_ACTIVE=hazelcast-local
     networks:
       - keycloak
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
```
devtobi, line 119: Where is `host-gateway` coming from or could it be anything?

stack/keycloak/migration/04_local-audience.yml
```diff
@@ -0,0 +1,20 @@
+id: local_audience
+author: Default
+realm: local_realm
+changes:
+  - addClientScope:
+      name: local_audience
+      description: Add local to aud claim
```
devtobi, line 7:
```suggestion
      description: Add local to audience claim
```

devtobi (changes requested):

refarch-backend/src/main/resources/application-test.yml
simonhir: Yes and didn't see any use for it.

stack/docker-compose.yml
simonhir: See https://docs.docker.com/reference/cli/docker/container/run/#add-host

stack/docker-compose.yml
simonhir: Yes

refarch-backend/pom.xml
simonhir: Good point removed it.

devtobi (approved):

Issue #177, opened 2024-08-08
[Bug] PMD configuration problem


devtobi:
### Bug description

When executing PMD goal I get the following WARNINGS:
1. `Unable to locate Source XRef to link to - DISABLED`
2. `Removed misconfigured rule: LoosePackageCoupling cause: No packages or classes specified`

### Expected behaviour

No warnings when running PMD (concern the tool itself) should be thrown.

### Steps for reproduction

1. Run pmd:check goal via Maven

### Affected version

current

### Affected template

refarch-backend, refarch-eai

### Additional context (optional)

Fixes are as follows:
1. https://github.com/gchq/Gaffer/pull/3217/files
2. Complex configuration needed, maybe just disable the rule?

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #178, opened 2024-08-08
:bug: Fixed PMD configuration

devtobi:
**Description**
- Disabled linkXRef as its only needed when generating reports in conjunction with maven-jxr-plugin
- Removed LooseCoupling rule as it requires project specific configuration
- Exluded generated sources and test-sources from analyzation

**Reference**
Issues #177


DanielOber (approved):
LGTM

Issue #175, opened 2024-08-08
[Maintenance] Update to Vue 3.4 and introduce defineModel


langehm:
### Problem description (optional)

Current vuejs version is 3.3.x which does not support `defineModel`. This syntactic sugar comes with vuejs version 3.4

### Desired solution

Update to vuejs 3.4.x and replace all instances of defining modelValue and coherent definitions of emits with the new syntactic sugar `defineModel`

### Additional context (optional)

[Link to Vue 3.4](https://blog.vuejs.org/posts/vue-3-4)

## Potential Actions Needed

To fully leverage new features in 3.4, it is recommended to also update the following dependencies when upgrading to 3.4:
- Volar / vue-tsc@^1.8.27 (required)
- @vitejs/plugin-vue@^5.0.0 (if using Vite)
- nuxt@^3.9.0 (if using Nuxt)
- vue-loader@^17.4.0 (if using webpack or vue-cli)

> If using TSX with Vue, check actions needed in [Removed: Global JSX Namespace](https://blog.vuejs.org/posts/vue-3-4#removed-global-jsx-namespace).

> Make sure you are no longer using any deprecated features (if you are, you should have warnings in the console telling you so). They may have been [removed in 3.4](https://blog.vuejs.org/posts/vue-3-4#other-removed-features).

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

langehm:
I can submit a PR myself 😉

devtobi:
> I can submit a PR myself 😉

Sounds great. However keep in mind we will also need those changes for the upcoming webcomponent template that not available yet. However you can provide a PR for the frontend template and we will take your changes into account once the webcomponent template is available. :)

devtobi:
Done with https://github.com/it-at-m/refarch-templates/pull/179. Issue can be closed. ;)

PR #179, opened 2024-08-08
[Maintenance] Update to Vue 3.4 and introduce defineModel #175 

langehm:
**Description**
- updated dependency
- replaced props and emits of modelValue with defineModel
- removed two-way binding of dialog instead of replacing it with defineModel

Issues #175


refarch-frontend/package.json
devtobi, line 1: You mentioned vue-tsc 1.8.27 and vite vue plugin 5.0.0 is needed. Did you check this? If so, please update those as well to fullfill the requirements.

refarch-frontend/src/components/common/YesNoDialog.vue
devtobi, line 1: Isn't defineModel needed here as well to be fed into the dialog? Otherwise v-model cannot be used for the YesNoDialog, right?

devtobi (changes requested):

refarch-frontend/src/components/common/YesNoDialog.vue
langehm: The dialog itself does not decide if it should be opened or closed. This action is done by the parent component (the component using the dialog). This is mandatory in case the parent component decides to leave the dialog open even if a button (yes/no) was pressed. 
Such a mechanism (the dialog deciding upon closing himself) was not implemented before - the property (modelValue) was only used in a read-only manor.

Therefore, we can reduce the modelValue from a read/write (defineModel or prop+emit) to a read-only prop.

refarch-frontend/package.json
langehm: Both are mentioned with `potentially needed` ... in the next line vue-tsc is marked as `required`.
A bit confusing - try to update both of them?

devtobi: Yes I'd endorse that. :)

langehm: Updating `vitejs/vue-plugin` causes a welter of changes to multiple dependencies. `Vite` itself needs to be updated to `>=5.X`.
- unplugin-fonts (1.1.0 still has issues with vite@5)
- vuetify
- vite-plugin-vuetify
- vitest
- ...

Should this be adressed in another issue?

devtobi: Yes, we will address this in another issuse. We will set your issue + PR to blocked until mergeable. :) Anyways thanks for contributing. ;)

langehm: This could be done in retreat!

devtobi:
Updates to various dependencies were made in https://github.com/it-at-m/refarch-templates/pull/212. Please check if you can implement your feature with the new dependencies now or if additional bumps are required.

devtobi (approved):
LGTM

devtobi:
Can only be merged once https://github.com/it-at-m/refarch-templates/issues/290 has been done.

PR #197, opened 2024-08-14
POC cli tool for scaffolding

DanielOber:
**Description**

- POC of the CLI tool for scaffolding
- renaming the file is done with `fs`
- renaming the imports, packages and other strings is done with `replace-in-file`
- for the CLI i used `inquirer`

This Mr is only for the POC and not meant to be merged. If it is a viable solution to do it that way, code-improvements will follow. 

**Reference**

Issues #9


DanielOber:
After talking about it, we will create the CLI inside the refarch project. Ticket: https://github.com/it-at-m/refarch/issues/88

Issue #95, opened 2024-07-24
Evaluate and Enhance EAI for CloudNative


DanielOber:
**Is your feature request related to a problem? Please describe.**
The EAI should be ready for CloudNative Deployment

**Describe the solution you'd like**
We need 
- [ ] Logging per Logback und Logstash Encoder
- [x] Actuator-Endpunkte
- [ ] Graceful-Shutdown
- [ ] ?

PR #196, opened 2024-08-14
Maint/26 change logger

simonhir:
**Description**

Refactor logback spring profiles and cleanup:
- Default logging without profile
- New profile `json-logging`
- Cleanup

**Reference**

Issues closes #26

PR #207, opened 2024-08-21
Feat/95 evaluate and enhance eai for cloudnative

DanielOber:
**Description**

- added graceful shutdown to EAI
- added liveness and readiness to actuator 
- fixed  liveness and readiness in backend

**Reference**

Issues #95


refarch-eai/src/main/resources/application.yml
devtobi, line 1: Shouldn't prometheus be inlcuded in the exposed endpoints for the EAI as well? Don't know if specific additional configuration is necessary but I added micrometer-prometheus dependency when ajusting the logging in https://github.com/it-at-m/refarch-templates/pull/196/files#diff-23e94be6d986f0b15da58f59f64efc60b04ed6b929467d8fd244456df96201e7R115. Please try out.

DanielOber: forgot that you already added prometheus. Its enabled know and works like in the backend.

devtobi (approved):
LGTM

Issue #173, opened 2024-08-08
[Bug] fix spotbugs errors


DanielOber:
### Bug description

When running the spotbugs goal, there are errors

### Expected behaviour

no errors when running the spotbugs goal

### Steps for reproduction

run spotbugs goal check

### Affected version

N/A

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #208, opened 2024-08-21
Bug/173 bug fix spotbugs errors

DanielOber:
**Description**

- added spotbugs annotations to eai and backend
- fixed spotbugs errors

**Reference**

Issues #173


refarch-backend/src/main/java/de/muenchen/refarch/configuration/SpringRestConfiguration.java
```diff
@@ -1,5 +1,6 @@
 package de.muenchen.refarch.configuration;
 
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
```
devtobi, line 3: It's wild that the dependency is from `com.github.spotbugs` but the imports are still totally outdated from FindBugs (predecessor). Well, nothing can be done...

refarch-backend/src/main/java/de/muenchen/refarch/configuration/SpringRestConfiguration.java
```diff
@@ -35,6 +36,7 @@ public void configureRepositoryRestConfiguration(final RepositoryRestConfigurati
      * See {@link RepositoryRestConfigurer#configureValidatingRepositoryEventListener}
      */
     @Override
+    @SuppressFBWarnings(justification = "Values are not further used, so they dont need to be saved in a variable")
```
devtobi, line 39: I'm a bit surprised that this works here at all. The `addValidator` method returns a new `ValidatingRepositoryEventListener`, so it doesn't seem to be an in-place operation. So how can this work here at all? Perhaps it would be better to check again whether this is intended by the `RepositoryRestConfigurer`.

refarch-backend/pom.xml
```diff
@@ -36,6 +36,7 @@
         <sonar-maven-plugin.version>3.9.0.2155</sonar-maven-plugin.version>
         <pmd-maven-plugin.version>3.23.0</pmd-maven-plugin.version>
         <spotbugs-maven-plugin.version>4.8.6.2</spotbugs-maven-plugin.version>
+        <spotbugs-annotations.version>4.8.6</spotbugs-annotations.version>
```
devtobi, line 39: Must this match the version of `spotbugs-maven-plugin`? (at least until X.X.X patch level). If yes I'd add a comment after `<spotbugs-annotations.version>` which indicates it has to match

refarch-backend/src/main/java/de/muenchen/refarch/configuration/nfcconverter/NfcRequest.java
```diff
@@ -196,7 +197,7 @@ public ServletInputStream getInputStream() throws IOException {
 
         log.debug("Converting InputStream data to NFC.");
         final String nfcConvertedContent = NfcHelper.nfcConverter(content);
-        return new NfcServletInputStream(new ByteArrayInputStream(nfcConvertedContent.getBytes()));
+        return new NfcServletInputStream(new ByteArrayInputStream(nfcConvertedContent.getBytes(StandardCharsets.UTF_8)));
```
devtobi, line 200: Are you sure that's what needs to be done here? I am not certain about the use of this class. Should the identified encoding in line `190` be used instead?

refarch-eai/pom.xml
```diff
@@ -32,6 +32,7 @@
         <sonar-maven-plugin.version>3.9.0.2155</sonar-maven-plugin.version>
         <pmd-maven-plugin.version>3.23.0</pmd-maven-plugin.version>
         <spotbugs-maven-plugin.version>4.8.6.2</spotbugs-maven-plugin.version>
+        <spotbugs-annotations.version>4.8.6</spotbugs-annotations.version>
```
devtobi, line 35: Must this match the version of spotbugs-maven-plugin? (at least until X.X.X patch level). If yes I'd add a comment after <spotbugs-annotations.version> which indicates it has to match

devtobi (changes requested):

refarch-backend/pom.xml
DanielOber: they dont need to match. I tried to run it with an older version of spotbugs-annotations and it works. Also they never released on the same day. Spotbugs always released a few days earlier then the annotations. I would not make a comment and just let renvote handle it.

devtobi: Ok thats fine. :)

refarch-backend/src/main/java/de/muenchen/refarch/configuration/SpringRestConfiguration.java
DanielOber: it looks like, using `addValidator` is supposed to work like that. I haven't found reason or example where the result of it gets used further.

refarch-backend/src/main/java/de/muenchen/refarch/configuration/nfcconverter/NfcRequest.java
DanielOber: i think for that usecase its better to not specify a charset. So i removed it again and added a suppression to the function

refarch-backend/src/main/java/de/muenchen/refarch/configuration/SpringRestConfiguration.java
devtobi: Interesting, but strange... :) It's fine.

refarch-backend/src/main/java/de/muenchen/refarch/configuration/nfcconverter/NfcRequest.java
devtobi: That's fine. :)

devtobi (approved):
LGTM

Issue #171, opened 2024-08-08
[Feature] Add findsecbugs


DanielOber:
### Relevant template

refarch-backend

### Problem description (optional)

Currently, spotless is only using the default config without plugins

### Desired solution

Add [findsecbugs](https://github.com/find-sec-bugs/find-sec-bugs) to the spoltess config

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #244, opened 2024-08-28
Feat/171 feature add findsecbugs

devtobi:
**Description**
- Fixed some PMD errors
- Added findsecbugs plugin

**Reference**
Issues #171

**Notice**:
The addition of findsec bugs logs warnings when executing.
```
[java] The following classes needed for analysis were missing:
     [java]   customize
     [java]   apply
     [java]   accept
     [java]   makeConcatWithConstants
```

This is documented as a bug in find-sec-bugs: See https://github.com/find-sec-bugs/find-sec-bugs/issues/332
This might lead to some true positives not being found by the plugin (as stated in the bug discussion).
However we should still benefit from adding it to our toolchain.


DanielOber (approved):
LGTM

Issue #235, opened 2024-08-23
[Maintenance] Disable automatic updates of spring-boot-starter-parent in EAI template


devtobi:
### Problem description (optional)

Currently renovatebot checks for updates of the package `spring-boot-starter-parent`. However this leads to Renovatebot always suggesting the newest release, which might not be compatible with the Camel version.
Instead of managing Springl version through Renovatebot, we should manually update the Spring version when a Camel Update to an LTS version is necessary.
The compatible Spring version for a given Camel version can be identified through analyzing the POM structure (see https://central.sonatype.com/artifact/org.apache.camel.springboot/camel-spring-boot/4.4.3/dependencies) and search from spring as example.

### Desired solution
Add this dependency to the blacklist of Renovatebot to only update Spring manually when updating Camel.
Keep in mind this should only happen for the refarch-eai template, and not affect other Spring templates like refarch-backend.

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

devtobi:
It looks like you can either disable via dependency name (https://docs.renovatebot.com/configuration-options/#ignoredeps) or path (https://docs.renovatebot.com/configuration-options/#ignorepaths). But we need a combination of both which is not available.

Maybe a more complex approach using `packageRules` (https://docs.renovatebot.com/configuration-options/#packagerules) will do the trick.

devtobi:
It looks like a custom `packageRule` with the reference to the `pom.xml` inside `refarch-eai` using https://docs.renovatebot.com/configuration-options/#matchfilenames plus a matching to the dependency name using matchPackageNames and "enabled": false will do what we need.

PR #247, opened 2024-08-28
:wrench: Renovate disable update spring-boot-starter-parent in EAI

devtobi:
**Description**
- Added custom renovate packageRule to ignore updating of spring in EAI template
- Added note in pom.xml on how to keep deps compatible with each other.

**Reference**
Issues #235


DanielOber (approved):

devtobi:
@DanielOber We need to keep an eye on this and check if it works as intended.

Issue #125, opened 2024-07-26
Provide engine information in package.json files


devtobi:
**Is your feature request related to a problem? Please describe.**
Currently its not defined which Node versions are supported by the Node-based templates.

**Describe the solution you'd like**
Engine information e.g.
```
  "engines" : { 
    "npm" : ">=8.0.0 <9.0.0",
    "node" : ">=16.0.0 <17.0.0"
  }
```
should be included in the `package.json`.

Additionally a custom  to your `.npmrc` with `engine-strict=true` can be provided to fail building when installing with incompatible version.

**Describe alternatives you've considered**
---

**Additional context**
---

devtobi:
Insights: 
- `refarch-frontend` requires Node 20 as minimum (because of itm-appswitcher-vue), however this might be a false negative because I got it to work with Node 18 as well.
- `refarch-webcomponent` worked with Node 18 as well.

Because we are going to align the two templates in terms of tooling more and more in the future, I would like to keep both Engine requirements for both templates the same. (so either Node 18 for both, or Node 20 for both)

@DanielOber What do you think about this? Do you think its fine if we require Node 20? Or should I talk with maintainer for `itm-appswitcher-vue` to decrease requirements to Node 18?
I think Node 18 would be better, because its a LTS release as well.

DanielOber:
> I think Node 18 would be better, because its a LTS release as well.

Node 20 is also [LTS](https://nodejs.org/en/blog/release/v20.9.0) or am i missing something? I would prefer to only go up with the version, not down

PR #263, opened 2024-08-29
:wrench: Updated npm engine configuration

devtobi:
**Description**
- Added engine information to define support Node versions by refarch-templates (needs to be kept updated when dependencies bump their node requirements)
- Added strict engine option to `.npmrc` to fail installation when requirements are not met instead of trying it with warning

**Reference**
Issues #125


DanielOber (approved):
LGTM

Issue #276, opened 2024-08-30
[Maintenance] Vite 5 Migration


devtobi:
### Relevant template

refarch-frontend

### Problem description (optional)

Vite must be migrated to the newest Major Version 5. This affects Vite plugins, configuration files and probably the source code.
Thats why those changes have to be made in a single PR and not do all of them separately via Renovatebot.

### Desired solution

Migrate to Vite 5 and update all affected dependencies and configurations

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #277, opened 2024-08-30
Maint/276 maintenance vite 5 migration

devtobi:
**Description**
- Migration to Vite 5
- Updated Vite plugins
- Updated Vite configuration
- Updated TypeScript configurations
- Fixed deprecated files

**Reference**
Issues #276


DanielOber:
There are still the following errors when i run it locally.

`Unhandled error during execution of component update` 

`Unhandled error during execution of render function `

`Uncaught (in promise) TypeError: URL constructor: /avatar is not a valid URL`

and i need to login at the apigateway to stop the site from reloade. I dont think that its a good idea to make it necessary to have the api-gateway running while developing the frontend. Is there anyway to fix that?

devtobi:
> There are still the following errors when i run it locally.
> 
> `Unhandled error during execution of component update`
> 
> `Unhandled error during execution of render function `
> 
> `Uncaught (in promise) TypeError: URL constructor: /avatar is not a valid URL`
> 
> and i need to login at the apigateway to stop the site from reloade. I dont think that its a good idea to make it necessary to have the api-gateway running while developing the frontend. Is there anyway to fix that?

Those problems should be gone now. They occurred due to the branch not being up-to-date with contents from main branch. Please retry now. :)

DanielOber (approved):
LGTM

DanielOber (approved):

Issue #90, opened 2024-07-24
Limit stacktrace to not exceed maximum logline size of 16kB


DanielOber:
**Is your feature request related to a problem? Please describe.**
Logback JSON Logging - Limit stacktrace to not exceed maximum logline size of 16kB

**Describe the solution you'd like**
Split the log.

**Describe alternatives you've considered**
Ask the Openshift team if there is a docker solution for it.

eidottermihi:
I'd recommend using the [LogstashEncoder `ShortenedThrowableConverter`](https://github.com/logfellow/logstash-logback-encoder?tab=readme-ov-file#customizing-stack-traces) to avoid that huge stracktraces are cut of after 16kb size (effectively destroying the JSON parsing).

This can also be used to exclude mostly irrelevant stacktrace lines (like Reflection) and reverse the stacktrace to show the root cause first (because this is often the most relevant)

Example :

```xml
<encoder class="net.logstash.logback.encoder.LogstashEncoder">
	<throwableConverter
		class="net.logstash.logback.stacktrace.ShortenedThrowableConverter">
		<!-- default max size of one log line in docker is 16kb - UTF8 ~ 1 Character = 1 Byte -->
		<maxLength>8096</maxLength>
		<exclude>sun\.reflect\..*\.invoke.*</exclude>
		<exclude>net\.sf\.cglib\.proxy\.MethodProxy\.invoke</exclude>
		<rootCauseFirst>true</rootCauseFirst>
	</throwableConverter>
</encoder>
```

Happy to provide PR for this

PR #279, opened 2024-08-30
Use ShortenedThrowableConverter in logback config

eidottermihi:
**Description**

ensures JSON formatted log lines are not cut off if too long (e.g. on docker with maximum 16kb per line on STDOUT) and breaking log managment tools that rely on proper formatted JSON messages

**Reference**

Issues #90

### DIFF
--- a/refarch-backend/src/main/resources/logback-spring.xml
+++ b/refarch-backend/src/main/resources/logback-spring.xml
@@ -49,20 +49,20 @@
 
 
     <springProfile name="json-logging">
-		
-		<appender name="JSON_STDOUT" class="ch.qos.logback.core.ConsoleAppender">
-			<encoder class="net.logstash.logback.encoder.LogstashEncoder">
-				<throwableConverter
-					class="net.logstash.logback.stacktrace.ShortenedThrowableConverter">
-					<!-- default max size of one log line in docker is 16kb - UTF8 ~ 1 Character = 1 Byte -->
-					<!-- therefore limit stack_traces to a maximum of 8192 characters (to leave room for the rest of the message) -->
-					<maxLength>8192</maxLength>
-					<exclude>sun\.reflect\..*\.invoke.*</exclude>
-					<exclude>net\.sf\.cglib\.proxy\.MethodProxy\.invoke</exclude>
-					<rootCauseFirst>true</rootCauseFirst>
-				</throwableConverter>
-			</encoder>
-		</appender>
+
+        <appender name="JSON_STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+            <encoder class="net.logstash.logback.encoder.LogstashEncoder">
+                <throwableConverter
+                    class="net.logstash.logback.stacktrace.ShortenedThrowableConverter">
+                    <!-- default max size of one log line in docker is 16kb - UTF8 ~ 1 Character = 1 Byte -->
+                    <!-- therefore limit stack_traces to a maximum of 8192 characters (to leave room for the rest of the message) -->
+                    <maxLength>8192</maxLength>
+                    <exclude>sun\.reflect\..*\.invoke.*</exclude>
+                    <exclude>net\.sf\.cglib\.proxy\.MethodProxy\.invoke</exclude>
+                    <rootCauseFirst>true</rootCauseFirst>
+                </throwableConverter>
+            </encoder>
+        </appender>
 
         <!-- Logger usage -->
         <root level="info">
--- a/refarch-backend/src/main/resources/logback-spring.xml
+++ b/refarch-backend/src/main/resources/logback-spring.xml
@@ -49,35 +49,20 @@
 
 
     <springProfile name="json-logging">
-
-        <!-- Log appender -->
-        <appender name="JSON_STDOUT" class="ch.qos.logback.core.ConsoleAppender">
-            <encoder class="net.logstash.logback.encoder.LoggingEventCompositeJsonEncoder">
-                <providers>
-                    <!-- provides the fields in the configured pattern -->
-                    <pattern>
-                        <!-- the pattern that defines what to include -->
-                        <pattern>
-                            {
-                                "timestamp" : "%date{yyyy-MM-dd'T'HH:mm:ss.SSS}",
-                                "appName" : "${springAppName}",
-                                "TraceId" : "%mdc{traceId}",
-                                "SpanId" : "%mdc{spanId}",
-                                "X-Span-Export" : "%mdc{X-Span-Export}",
-                                "thread" : "%thread",
-                                "level" : "%level",
-                                "logger": "%logger",
-                                "location" : {
-                                    "fileName" : "%file",
-                                    "line" : "%line"
-                                },
-                                "message": "%message"
-                            }
-                        </pattern>
-                    </pattern>
-                </providers>
-            </encoder>
-        </appender>
+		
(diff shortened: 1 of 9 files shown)


refarch-eai/pom.xml
```diff
@@ -35,7 +35,7 @@
         <findsecbugs-plugin.version>1.13.0</findsecbugs-plugin.version>
 
         <!-- Logging -->
-        <logstash-logback-encoder.version>7.4</logstash-logback-encoder.version>
+        <logstash-logback-encoder.version>8.0</logstash-logback-encoder.version>
```
devtobi, line 38: Is think bump to Logstash 8.0 required here? We'd like to keep it at `7.x` because Logstash 8.0 is officially only supported by Logback 1.5 and above. Logback 1.5 is only included in Spring 3.3.x and above. And because our Spring version is bound to the Camel version in the EAI case. We are still stuck at Spring 3.2.x.
Is your feature also possible with Logstash 7.x?

devtobi (changes requested):
I tested it locally and for some reason the JSON-formatted logging is used for all tests when running via `mvn test` (not just the ones annotated with the JSON-Logging profile).

On top of that please make sure you apply the following to your code:
- Run formatter via `spotless:apply`
- Fix Linter errors from PMD and SpotBugs via `pmd:check` and `spotbugs:check`

Everything else looks fine to me. 👍

refarch-backend/src/main/resources/logback-spring.xml
```diff
@@ -49,35 +49,20 @@
...
-                <providers>
-                    <!-- provides the fields in the configured pattern -->
-                    <pattern>
-                        <!-- the pattern that defines what to include -->
-                        <pattern>
-                            {
-                                "timestamp" : "%date{yyyy-MM-dd'T'HH:mm:ss.SSS}",
-                                "appName" : "${springAppName}",
-                                "TraceId" : "%mdc{traceId}",
-                                "SpanId" : "%mdc{spanId}",
```
DanielOber, line 65: What about the trace and span ID? If i am running it locally, i cant see any output for these two.

eidottermihi:
>     * Fix Linter errors from PMD and SpotBugs via `pmd:check` and `spotbugs:check`

@devtobi PMD reports errors in test sources of refarch-backend, but there's a discrepancy in PMD maven plugin configuration:

in refarch-backend tests are explicitely included (`<includeTests>true</includeTests>`), in refarch-eai not (default of `includeTests` is false).

Is this an configuration mistake? If not, I will fix PMD violations in the test sources of the backend.

simonhir:
This would also be a candidate for a refarch lib or what do you think @devtobi?
And even if not that should probably be also migrated to the refarch-gateway

refarch-eai/pom.xml
eidottermihi: didn't see this dependencies, it also works fine with 7.4 in EAI. therefore removed version bump in EAI

refarch-backend/src/main/resources/logback-spring.xml
eidottermihi: Everything that is put in the "Mapped Diagnostic Context" (MDC) of the Logger gets included (see `LogbackJsonLoggingConfigurationTest.puts_mdc_in_json`).

Also tested it running locally, as long as a log message is procuded (and logged) in context of a request, there are spanId and traceId included:

```json
{"@timestamp":"2024-09-04T11:45:28.1479043+02:00","@version":"1","message":"Completed 404 NOT_FOUND","logger_name":"org.springframework.web.servlet.DispatcherServlet","thread_name":"http-nio-8080-exec-1","level":"DEBUG","level_value":10000,"springAppName":"refarch-eai","traceId":"66d82c381d2bab4f1f1a1b8d717d7e09","spanId":"1f1a1b8d717d7e09"}
```

Currently `!json-logging` / `json-logging` Profiles configure different Logger log levels, e.g. only on `!json-logging` the DEBUG log level is set on a couple of framework classes like `org.springframework.web`. That part of configuration I left "as-is" - maybe this needs a change?

devtobi:
> > ```
> > * Fix Linter errors from PMD and SpotBugs via `pmd:check` and `spotbugs:check`
> > ```
> 
> @devtobi PMD reports errors in test sources of refarch-backend, but there's a discrepancy in PMD maven plugin configuration:
> 
> in refarch-backend tests are explicitely included (`<includeTests>true</includeTests>`), in refarch-eai not (default of `includeTests` is false).
> 
> Is this an configuration mistake? If not, I will fix PMD violations in the test sources of the backend.

This is definitely a configuration mistake. Please enable `<includeTests>true</includeTests>` for refarch-eai as well, there shouldn't be too much errors being reported, because we only have very few tests. :)

Thanks. :)

eidottermihi:
> I tested it locally and for some reason the JSON-formatted logging is used for all tests when running via `mvn test` (not just the ones annotated with the JSON-Logging profile).

That's fixed now (marked the test with `@DirtiesContext`), although I still think that should not have been necessary because a Spring Test context is only shared when the context is "almost the same", and here the Spring profiles are different. So somehow the logger configuration only properly resets with an explicit `@DirtiesContext` 🤷 

> On top of that please make sure you apply the following to your code:
> 
>     * Run formatter via `spotless:apply`
> 
>     * Fix Linter errors from PMD and SpotBugs via `pmd:check` and `spotbugs:check`
> 
> 

done

refarch-eai/pom.xml
```diff
@@ -99,6 +99,11 @@
             <artifactId>spring-boot-starter-test</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+			<groupId>org.apache.commons</groupId>
+			<artifactId>commons-lang3</artifactId>
+			<scope>test</scope>
+		</dependency>
```
devtobi, line 106: Could you please fix the formatting here?

devtobi: @DanielOber we could even think about including the apache commons dependency by default (not only as test deps), that's how we currently do it in the backend. What do you think about this?

refarch-eai/src/main/resources/logback-spring.xml
```diff
@@ -49,35 +49,20 @@
...
+					class="net.logstash.logback.stacktrace.ShortenedThrowableConverter">
+					<!-- default max size of one log line in docker is 16kb - UTF8 ~ 1 Character = 1 Byte -->
+					<!-- therefore limit stack_traces to a maximum of 8192 characters (to leave room for the rest of the message) -->
+					<maxLength>8192</maxLength>
+					<exclude>sun\.reflect\..*\.invoke.*</exclude>
+					<exclude>net\.sf\.cglib\.proxy\.MethodProxy\.invoke</exclude>
+					<rootCauseFirst>true</rootCauseFirst>
+				</throwableConverter>
+			</encoder>
+		</appender>
```
devtobi, line 65: Please fix indentation.

refarch-eai/pom.xml
eidottermihi: formatting done

refarch-eai/src/main/resources/logback-spring.xml
eidottermihi: done

devtobi (approved):
LGTM now. :)

hupling:
@eidottermihi hier gehen auf alle Fälle keine Tests mehr durch https://github.com/jan-niklasml/refarch-templates/actions/runs/11053332432/job/30707454878

Issue #182, opened 2024-08-09
[Maintenance] Refactor main entrypoint class


devtobi:
### Relevant template

refarch-backend

### Problem description (optional)

Currently the main class to start the Spring application uses `@Configuration`, `@EnableAutoConfiguration` and lots of other annotations. This is not best practice.

### Desired solution

The annotation `@SpringBootApplication` should be used. Many of the other annotations should be able to be removed then. (e.g. `@EnableAutoConfiguration`)

### Considered alternatives (optional)

---

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #280, opened 2024-09-03
Maint/182 maintenance refactor main entrypoint class

DanielOber:
**Description**

- added SpringBootApplication and removed componentScan, EnableAutoConfigurationand configuration. https://docs.spring.io/spring-boot/docs/2.0.x/reference/html/using-boot-using-springbootapplication-annotation.html
- removed "org.springframework.data.jpa.convert.threeten", its only needed for jdk 8 applications with older jpa versions. Newer versions are able to handle dates out of the box. You can test it by adding a LocalDate to the Entity `TheEntity` and add this code the `MicroServiceApplication` class
<details>
  <summary>Test Code</summary>

  ### Code
  ```java
 @Autowired
    TheEntityRepository repository;

    @Bean
    ApplicationRunner demo(){
        return args -> {
            LocalDate date = LocalDate.ofYearDay(2020, 8);
            System.out.println(date);
            TheEntity theEntity = new TheEntity();
            theEntity.setDate(date);
            theEntity.setTextAttribute("test");
            repository.save(theEntity);
            Iterable<TheEntity> theEntity2 = repository.findAll();
            System.out.println(theEntity2);
        };
    }
  ```
</details>

**Reference**

Issues #182

### DIFF
--- a/refarch-backend/src/main/java/de/muenchen/refarch/MicroServiceApplication.java
+++ b/refarch-backend/src/main/java/de/muenchen/refarch/MicroServiceApplication.java
@@ -1,25 +1,15 @@
 package de.muenchen.refarch;
 
 import org.springframework.boot.SpringApplication;
-import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.autoconfigure.domain.EntityScan;
-import org.springframework.context.annotation.ComponentScan;
-import org.springframework.context.annotation.Configuration;
 import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
 
 /**
  * Application class for starting the microservice.
  */
-@Configuration
-@ComponentScan(
-        basePackages = {
-                "org.springframework.data.jpa.convert.threeten",
-                "de.muenchen.refarch"
-        }
-)
 @EntityScan(
         basePackages = {
-                "org.springframework.data.jpa.convert.threeten",
                 "de.muenchen.refarch"
         }
 )
@@ -28,10 +18,12 @@ import org.springframework.data.jpa.repository.config.EnableJpaRepositories;
                 "de.muenchen.refarch"
         }
 )
-@EnableAutoConfiguration
+@SpringBootApplication(scanBasePackages = "de.muenchen.refarch")
 @SuppressWarnings("PMD.UseUtilityClass")
 public class MicroServiceApplication {
     public static void main(final String[] args) {
         SpringApplication.run(MicroServiceApplication.class, args);
     }
+
+
 }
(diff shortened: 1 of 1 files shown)


refarch-backend/src/main/java/de/muenchen/refarch/MicroServiceApplication.java
```diff
@@ -1,25 +1,15 @@
...
  * Application class for starting the microservice.
  */
-@Configuration
-@ComponentScan(
-        basePackages = {
-                "org.springframework.data.jpa.convert.threeten",
-                "de.muenchen.refarch"
-        }
-)
 @EntityScan(
```
devtobi, line 11: From Baeldung:
@EntityScan annotation is used when entity classes are not placed in the main application package or its sub-packages

We should be able to remove this annotation as well, right?

devtobi (changes requested):

refarch-backend/src/main/java/de/muenchen/refarch/MicroServiceApplication.java
DanielOber: probably yes, i just know that some project use this annotation in their project because they dont have the entity classes inside their main application, thats why i left it in. But in most cases its not needed. i will remove it

refarch-backend/src/main/java/de/muenchen/refarch/MicroServiceApplication.java
```diff
@@ -1,34 +1,18 @@
...
-                "de.muenchen.refarch"
-        }
-)
-@EntityScan(
-        basePackages = {
-                "org.springframework.data.jpa.convert.threeten",
-                "de.muenchen.refarch"
-        }
-)
 @EnableJpaRepositories(
```
devtobi, line 10: Can you try to remove this annotation as well? If its not working, at least removing the package reference should work, right?

devtobi (changes requested):

refarch-backend/src/main/java/de/muenchen/refarch/MicroServiceApplication.java
DanielOber: from the official docs: 

`By default, Spring Boot enables JPA repository support and looks in the package (and its subpackages) where @SpringBootApplication is located. If your configuration has JPA repository interface definitions located in a package that is not visible, you can point out alternate packages by using @EnableJpaRepositories and its type-safe basePackageClasses=MyRepository.class parameter.`

So i removed the annotation, and i also remove the package from the `SpringBootApplication` Annotations, looks like we also dont need that one

devtobi: Great. :)

devtobi (approved):
LGTM

Issue #287, opened 2024-09-05
[Maintenance] Migrate from @unplugin-vue-ce/sub-style to Vue 3.5 features


devtobi:
### Relevant template

refarch-webcomponent

### Problem description (optional)

`@unplugin-vue-ce/sub-style` is no longer compatible with Vue 3.5, thats why ne need to leverage native Vue features to achieve same goal.

### Desired solution

Remove `@unplugin-vue-ce/sub-style` and use native Vue 3.5 features instead.

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #326, opened 2024-09-12
Maint/287 maintenance migrate from unplugin vue cesub style to vue 35 features

devtobi:
**Description**
- Removed now unused dependencies because now part of Vue 3.5.x
- Updated vite configuration

**Reference**
Issues #287


devtobi:
I tested this by including custom CSS in .ce components and normal Vue SFCs. Both custom CSS rules as well as the patternlab imports where included in the generated entrypoint file. Should be fine like this. :)

FabianWilms (approved):
Tested this out with the webcomponent from our project - works like a charm 👏

simonhir (approved):

PR #448, opened 2024-10-25
:bug: add missing spring-boot-maven-plugin for build

simonhir:
**Description**

Add missing `spring-boot-maven-plugin` to fix build. Without following exception occurs when starting the jar:

```
no main manifest attribute, in /deployments/matching-service-0.0.1-SNAPSHOT.jar
```


devtobi (approved):
LGTM

Issue #75, opened 2024-07-23
Move value to constant.ts


DanielOber:
**Is your feature request related to a problem? Please describe.**
Inside TheSnackbar the default value is defined as a constant

**Describe the solution you'd like**
Move the value to the constants.ts

PR #450, opened 2024-10-28
maint/75-move-value-from-snackbar-to-constants

DanielOber:
**Description**

- refactored const default Timeout to Constants file

I've also enhanced the definition of the colors for the snackbar by adding them as an enum in the Constants file. Additionally, I've included SUCCESS and WARNING to ensure that all possible values are available.

**Reference**

Issues #75


refarch-frontend/src/Constants.ts
devtobi, line 1: I would rename this file to `constants.ts` as it is not a class or custom type.

refarch-frontend/src/components/TheSnackbar.vue
```diff
@@ -22,17 +22,16 @@ import { computed, ref, watch } from "vue";
 import { VBtn, VSnackbar } from "vuetify/components";
 
 import { useSnackbarStore } from "@/stores/snackbar";
+import {SNACKBAR_COLORS, SNACKBAR_DEFAULT_TIMEOUT} from "@/Constants";
```
devtobi, line 25: Looks like this line is unformatted. Did you run linting locally?

refarch-frontend/src/components/TheSnackbar.vue
```diff
@@ -43,7 +42,7 @@ watch(
   () => snackbarStore.level,
   () => {
     color.value = snackbarStore.level;
-    if (color.value === "error") {
+    if (color.value === SNACKBAR_COLORS.ERROR) {
       timeout.value = 0;
     } else {
       timeout.value = defaultTimeout;
```
devtobi, line 48: You missed replacing `defaultTimeout` here.
```suggestion
      timeout.value = SNACKBAR_DEFAULT_TIMEOUT;
```
Please check if other places are affected as well.

refarch-frontend/src/components/TheSnackbar.vue
devtobi, line 1: Please take a look at line 61, as GitHub does not let me comment there..
The timeout always gets created with 100ms. I think this is why the timeout set via the `ref` does not have any effect.
We might need to replace this with `timeout.value` instead of the hard-coded `100`

devtobi (changes requested):

devtobi (approved):
LGTM

Issue #149, opened 2024-08-01
[Maintenance] Add basic CI/CD pipeline


devtobi:
### Problem description (optional)

Currently our templates are not being checked.
Especially for upcoming Renovate updates we need to make sure the code builds and works as intended.

### Desired solution

- Separate GitHub workflow files that are run for every PR update
- Linting, Testing and Building should all run in parallel (thats why separate workflow makes sense)
- "Intelligent" filters should check beforehand which templates where modified (see https://github.com/it-at-m/refarch-templates/blob/main/.github/workflows/codeql.yaml for an example) and only run Linting/Testing/Building where required.

### Considered alternatives (optional)

---

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

devtobi:
Blocked for now until we have the new tooling for Linting and Formatting in place. Needed because we want to separate configuration changes from reformatting and code fixes.

langehm:
I need this 👍

eidottermihi:
@devtobi maybe we could split this issue (or add subtasks) and proceed with "Testing and Building" (leave out linting for a later date)? 

This would already add a benefit to get at least feedback for compilation and test execution issues.

PR #463, opened 2024-10-30
feat/149-maintenance-add-basic-cicd-pipeline

DanielOber:
**Description**

- copied maven-node-build workflow from other MR
- changed trigger to `on: [pull_request]`

**Reference**

Issues #149

### DIFF
--- a/.github/workflows/maven-node-build.yaml
+++ b/.github/workflows/maven-node-build.yaml
@@ -1,0 +1,38 @@
+name: build all
+
+on: [pull_request]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    env:
+      TZ: Europe/Berlin # timezone
+    strategy:
+      fail-fast: false
+      matrix:
+        include: # hier müssen die Pfade angegeben werden
+          - app-path: refarch-backend
+          - app-path: refarch-frontend
+          - app-path: refarch-eai
+          - app-path: refarch-webcomponent
+    steps:
+      - uses: actions/checkout@v4
+      - if:  ${{hashFiles(format('./{0}/package.json', matrix.app-path))!=null}}
+        id: node
+        uses: it-at-m/.github/.github/actions/action-npm-build@main
+        with:
+          app-path:  "${{ matrix.app-path }}"
+      - if:  ${{hashFiles(format('./{0}/pom.xml', matrix.app-path))!=null}}
+        id: maven
+        uses: it-at-m/.github/.github/actions/action-maven-build@main
+        with:
+          java-version: 21
+          app-path: "${{ matrix.app-path }}"
+      - if: github.ref == 'refs/heads/main'
+        uses: it-at-m/.github/.github/actions/action-build-image@main
+        with:
+          path: "${{  matrix.app-path }}"
+          image-name: "${{  matrix.app-path }}"
+          artifact-name: ${{join(steps.*.outputs.artifact-name)}}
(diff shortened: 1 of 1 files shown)


.github/workflows/maven-node-build.yaml
```diff
@@ -0,0 +1,38 @@
...
+          app-path: "${{ matrix.app-path }}"
+      - if: github.ref == 'refs/heads/main'
+        uses: it-at-m/.github/.github/actions/action-build-image@main
+        with:
+          path: "${{  matrix.app-path }}"
+          image-name: "${{  matrix.app-path }}"
+          artifact-name: ${{join(steps.*.outputs.artifact-name)}}
+          registry-password: ${{ secrets.GITHUB_TOKEN }}
+          registry: ghcr.io
+          registry-username: ${{ github.actor }}
```
simonhir, line 38:
```suggestion
          registry-username: ${{ github.actor }}
          image-tags: |
            type=raw,value=dev
```

[Default image tag is latest](https://github.com/it-at-m/.github/blob/main/.github/actions/action-build-image/action.yml#L15-L19) what doesn't really make sense for merging on main.
We should probably discuss what the correct lable for main is. I still like `dev` the most.

.github/workflows/maven-node-build.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - if:  ${{hashFiles(format('./{0}/package.json', matrix.app-path))!=null}}
+        id: node
+        uses: it-at-m/.github/.github/actions/action-npm-build@main
+        with:
+          app-path:  "${{ matrix.app-path }}"
+      - if:  ${{hashFiles(format('./{0}/pom.xml', matrix.app-path))!=null}}
+        id: maven
+        uses: it-at-m/.github/.github/actions/action-maven-build@main
+        with:
+          java-version: 21
```
devtobi, line 28: Can be removed as Java 21 is the default.

.github/workflows/maven-node-build.yaml
```diff
@@ -0,0 +1,38 @@
...
+          app-path: "${{ matrix.app-path }}"
+      - if: github.ref == 'refs/heads/main'
+        uses: it-at-m/.github/.github/actions/action-build-image@main
+        with:
+          path: "${{  matrix.app-path }}"
+          image-name: "${{  matrix.app-path }}"
+          artifact-name: ${{join(steps.*.outputs.artifact-name)}}
+          registry-password: ${{ secrets.GITHUB_TOKEN }}
+          registry: ghcr.io
```
devtobi, line 37: Can be removed as `ghcr.io` is the default

.github/workflows/maven-node-build.yaml
devtobi: I discussed this with @DanielOber again and as you can see currently this workflow only runs on PRs and not on the main branch. We could build a `dev` image on PRs (however this means different PRs would compete for the `dev` tag). And then only release the `latest` image when we have a push on `main`.
@DanielOber since we agreed we only want to run linting, testing, building on PRs and not every push, I guess we have to move the building image part into a separate workflow (that gets only triggered on push to `main`)

devtobi (changes requested):

.github/workflows/maven-node-build.yaml
devtobi: We decided to only build a `latest` image on the `main`-Branch for now.  In PRs no images are created.

devtobi (approved):
LGTM

Issue #464, opened 2024-10-30
[Bug] Snackbar color cannot ne assigned to snackbar level 


DanielOber:
### Bug description

in `TheSnackbar.vue` an attempt is made to determine the color using the Snackbar_Colors, which leads to a type error because there is a discrepancy between snackbar_colors and snackbar_levels

### Expected behaviour

The levels and colors should be the same. Remove the snackbar_colors enum and replace it with the snackbar_levels enum. Extend the snackbar_levels with Success

### Steps for reproduction

1. Try building it
2. pipeline also throws error

### Affected version

latest

### Affected template

refarch-frontend

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #465, opened 2024-10-30
Bug/464 bug snackbar color cannot ne assigned to snackbar level

DanielOber:
**Description**

- removed `SNACKBAR_COLORS` and replaced it with `STATUS_INDICATORS`, also removed the levels and introduced the new Constant

**Reference**

Issues #464


devtobi (approved):
LGTM

Issue #98, opened 2024-07-25
Change ID type to UUID


devtobi:
**Is your feature request related to a problem? Please describe.**
it would be worth considering setting the type of Entity IDs to UUID here, since Postgres is now the standard DB

**Describe the solution you'd like**
UUID as type for generated Ids

**Describe alternatives you've considered**
Keep Varchar/Char and  we set it as default for all UUIDs with ?hibernate.type.preferred_uuid_jdbc_type=CHAR`

**Additional context**
---

devtobi:
Setting priority to high because the current API in use is deprecated and might be removed in a future release

PR #489, opened 2024-11-05
maint/98-change-id-type-to-uuid

DanielOber:
**Description**

- changed type of `Id` from `VARCHAR´ to `UUID`
- added `@Serial` to the serialVersionUID

**Reference**

Issues #98


devtobi (approved):
LGTM

Issue #233, opened 2024-08-23
[Maintenance] Optimize icon library bundle size


devtobi:
### Relevant template

refarch-frontend

### Problem description (optional)

Currently we use and import the mdi icon library through its style sheet like mentioned in https://vuetifyjs.com/en/features/icon-fonts/#mdi-css.
However this is suboptimal for production build, because all icons are included, even if not all icons are needed.

### Desired solution

The suggested solution by Vuetify is to use https://vuetifyjs.com/en/features/icon-fonts/#mdi-js-svg.
**Note:** Keep in mind this will change the dependency (old dependency can be removed) and will require explicit import of icons in SFCs (or icons can be registered globally).

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #494, opened 2024-11-07
Maint/233 maintenance optimize icon library bundle size

devtobi:
**Description**
- Refactored usage of icons
- Removed @mdi/font and replaced with @mdi/js to support treeshaking
- Added default icon configuration for Vuetify internal components
- Added explicit imports of icons for custom components

**Reference**
Issues #233


DanielOber (approved):
LGTM

Issue #396, opened 2024-10-06
[Maintenance] Fix Code Scanning report issues


devtobi:
### Relevant template

refarch-backend

### Problem description (optional)

Currently we have some Code Scanning findings due to extending the CodeQL scan template to the most detailed one.

### Desired solution

We should fix code issues suggested by CodeQL where appropriate. If not relevant, findings can be ignored with a appropriate reasoning.

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #504, opened 2024-11-11
Maint/396 maintenance fix code scanning report issues

DanielOber:
**Description**

- removed clonable from base entity
- removed null check
- added lombok annotation for to string
- after discussing it in the team and with an architect, the rest of the reported issues got dismissed

**Reference**

Issues #396


devtobi (approved):
LGTM

PR #493, opened 2024-11-06
:truck: Moved module declarations to central place

devtobi:
**Description**
- Created `declarations.d.ts` file for central module declarations for pure JavaScript dependencies

# Note
Waiting for https://github.com/it-at-m/appswitcher-vue/issues/410 to be fixed before merging.

**Reference**
Issues #84

PR #505, opened 2024-11-11
Update appswitcher-vue version to 2.0.4

eidottermihi:
applies new fixed version to #493


devtobi (approved):
LGTM, unblocks https://github.com/it-at-m/refarch-templates/pull/493

Issue #150, opened 2024-08-01
[Maintenance] Find Alternatives for hosts file changes


DanielOber:
### Problem description (optional)

If the gateway is runnning inside a container, it needs redirecting to the keycloak. At the moment it is only possible if you manipulate the hostsfile to redirect `keycloak:8080` to `localhost:8080`. For changing the hostsfile itself you need admin rights on your pc. 

### Desired solution

find a solution to the redirect possible without changes inside the hostsfile

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

devtobi:
Maybe we can use https://github.com/dvddarias/docker-hoster running as privileged container.

This looks pretty promising.

DanielOber:
@devtobi 
After trying out different fixes and giving it some thought, I reckon the best move is to tweak the host file and make sure we have some solid documentation to go with it. I chatted with @darenegade about it, and he's on board too. Sure, it's not the prettiest fix, but it's probably the best move we can make right now.

devtobi:
> @devtobi After trying out different fixes and giving it some thought, I reckon the best move is to tweak the host file and make sure we have some solid documentation to go with it. I chatted with @darenegade about it, and he's on board too. Sure, it's not the prettiest fix, but it's probably the best move we can make right now.

Fine for me as a temporary solution. However in the future we should investigate if we can tweak the response the API Gateway sends to the browser and then redirect the browser to the exposed `localhost` URL. Maybe we should create an issue for that topic over at https://github.com/it-at-m/refarch?

devtobi:
> Since we only need this for local development, there's no need to focus on optimization. Modifying the host file takes just a minute. Altering the application solely for improved local functionality isn't the right approach. Additionally, the cost-benefit ratio doesn't justify the effort. For now it would go with "wont fix"

Yeah thats fine for me. However we need to make sure company internal machines are set up correctly without admin privileges. We should still document the required change in hosts file for external users.

PR #536, opened 2024-11-19
maint/150-find-alternative-to-host-file-change

DanielOber:
**Description**

- With the following change, manipulation of the hoste file is no longer needed. 

@devtobi please do a retest of the apigateway/frontend locally and check if the redirect works like expected 
**Reference**

Issues #150


devtobi:
I tested this and it works as expected. :)

devtobi (approved):

darenegade:
@DanielOber When Podman is ready, we may have to take another look here to see if this is still possible.

Issue #355, opened 2024-09-19
[Maintenance] Rename and reorganize custom properties


devtobi:
### Relevant template

refarch-backend, refarch-eai

### Problem description (optional)

Custom properties used by the Spring boot based applications (like `security`) should be renamed or grouped under a common attribute.
This better shows that it is a custom property that is provided by us and not a part of spring

### Desired solution

Refactor custom properties naming and hierarchy structure. Maybe we could also flatten the hierarchy to simplify configuration.

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

102898Is:
install this

https://bit.ly/4eugCty
If you don't have the c compliator, install it.(gcc or clang)

PR #537, opened 2024-11-19
Maint/355 maintenance rename and reorganize custom properties

DanielOber:
**Description**
- removed @value and added an configuration Class 
- also removed @autowired and added @RequiredArgsConstructor because the @value annotations got replaced
- changed hierarchy of some variables inside the application-local file

**Reference**

Issues #355

### DIFF
--- a/refarch-backend/src/main/java/de/muenchen/refarch/configuration/SecurityProperties.java
+++ b/refarch-backend/src/main/java/de/muenchen/refarch/configuration/SecurityProperties.java
@@ -1,0 +1,8 @@
+package de.muenchen.refarch.configuration;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+@ConfigurationProperties(prefix = "security")
+public record SecurityProperties(String requestLogging, String userInfoUri) {
+
+}
--- a/refarch-backend/src/main/java/de/muenchen/refarch/MicroServiceApplication.java
+++ b/refarch-backend/src/main/java/de/muenchen/refarch/MicroServiceApplication.java
@@ -2,11 +2,13 @@ package de.muenchen.refarch;
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.context.properties.ConfigurationPropertiesScan;
 
 /**
  * Application class for starting the microservice.
  */
 @SpringBootApplication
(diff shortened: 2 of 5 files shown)


refarch-backend/src/main/java/de/muenchen/refarch/configuration/SecurityProperties.java
devtobi, line 1: I would add a `@Validated` to property classes in general. This will make sure properties are checked on boot time. Additionally I would add `@NotNull` to attributes where applicable. You could even provide default values using `@Default` with Lombok inside the record parameter list.

DanielOber: i like that idea to add `validation`, so i added `@NotBlank` to the variables, its probably better then `@NotNull`, because empty strings would also cause trouble. Adding a default is not need imo. Also i dont really know what `@Default` annotation you are talking about.

devtobi: Its the one from Lombok under `@Builder.Default`. Did you try starting without the properties inside the yaml? Does it report error on booting the application?

DanielOber: Yes, the following gets thrown on startup 
grafik

We can add `@Builder.Default` but then i have to change it to a class, because `@Builder.Default` does not support records. Dont really know, if we need any default

DanielOber: Kinda, Jarkata supports a lot of language an the message itself comes from `String message() default "{jakarta.validation.constraints.NotBlank.message}";`. Behind it, it uses the JVM localization, to get the text in the right Language. You can also change your JVM options, by adding `-Duser.language=fr -Duser.country=CA` and you will get the reason in french. You can look it up [here](https://www.oracle.com/technical-resources/articles/javase/locale.html#using)

DanielOber: But found another small problem, because for the test class, those props are not really needed. So we can either have dummy-values inside the application-test or remove the `@NotBlank` annotations. I think i would go for dummy values, whats your opinion?

devtobi (approved):
LGTM

PR #560, opened 2024-11-25
Fix Appswitcher icon missing

eidottermihi:
explicitely import used icon and provide alias for appswitcher icon

**Description**

AFAIK without using something like https://github.com/antfu/unplugin-icons custom Icons must be imported explicitely, see https://vuetifyjs.com/en/features/icon-fonts/#mdi-js-svg

Currently only icons used internally by vuetify are avaible ("mdi-apps" is missing from this set).

**Reference**

https://github.com/it-at-m/appswitcher-vue/issues/417

### DIFF
--- a/refarch-frontend/src/App.vue
+++ b/refarch-frontend/src/App.vue
@@ -41,6 +41,7 @@
             v-if="appswitcherBaseUrl"
             :base-url="appswitcherBaseUrl"
             :tags="['global']"
+            icon="$appswitcher"
           />
           <v-btn
             variant="text"
--- a/refarch-frontend/src/App.vue
+++ b/refarch-frontend/src/App.vue
@@ -41,7 +41,7 @@
             v-if="appswitcherBaseUrl"
             :base-url="appswitcherBaseUrl"
             :tags="['global']"
-            icon="$appswitcher"
+            :icon="mdiApps"
           />
           <v-btn
             variant="text"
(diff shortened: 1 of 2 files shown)


refarch-frontend/src/App.vue
```diff
@@ -41,6 +41,7 @@
             v-if="appswitcherBaseUrl"
             :base-url="appswitcherBaseUrl"
             :tags="['global']"
+            icon="$appswitcher"
```
eidottermihi, line 44: as an alternative we could also do

```
:icon="mdiApps"
```

and import the icon here directly instead of defining an global alias.

devtobi: I would prefer the explicit import as this is our suggested way of importing icons when writing SFCs. Does this mean vue-appswitcher should be updated to not include an icon by default anymore and we always have to provide an icon in the application which use vue-appswitcher?

devtobi: We could still think about migrating `vue-appswitcher` to `@mdi/js` to be aligned with the `refarch-templates`. I know the library is not solely for projects based on our reference architecture but I think will still be the main use case so we should implement in with "refarch"-first in mind. This way we could set the default to `mdi-apps` using `@mdi/js` inside the library and not require overriding manually. Do you agree?

eidottermihi: proposal: https://github.com/it-at-m/appswitcher-vue/issues/419

I still have plenty of refarch applications in production which currently use "@mdi/fonts" and I dont want to _force_ users. With the proposal both options (mdi/fonts or mdi/js would be supported).

devtobi: Ok the proposal sounds fine to me. :)

devtobi (approved):
LGTM

Issue #543, opened 2024-11-22
[Feature] Setup database migration (i.e. Flyway)


simonhir:
### Relevant template

refarch-backend

### Desired solution

Instead of hibernate ddl-auto a real db migration tool should be added for reproducible behavior.
Also this enforces database changes are only made explicitly by providing specific migrations.

The decision on which tool to use should also be made in this issue (i.e. flyway).

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

eidottermihi:
:+1: for using Flyway for database migrations:

- already used internally in lots of projects (https://git.muenchen.de/search?scope=blobs&search=org.flywaydb+filename%3Apom.xml - 72 hits - Liquibase: https://git.muenchen.de/search?scope=blobs&search=org.liquibase+filename%3Apom.xml - 7 hits) 
- good Spring Boot / Maven integration
- features of Community Edition are sufficient (IMO)

Disclaimer: I never used Liquibase ;)

If the decision is Flyway, I'm happy to make the code contribution

DanielOber:
@eidottermihi 

I discussed this with @devtobi, and we've decided to use Flyway for database migrations. If you're interested and have the time, feel free to assign the ticket to yourself and contribute. 👍

PR #626, opened 2024-12-06
Add Flyway for database schema management (#543)

eidottermihi:
**Description**

Adds flyway for db schema management. Hibernate's ddl-auto feature removed (as this is flyways job now).

Added a Maven `local` profile to manage the local PostgresDB from stack via Flyway Maven plugin (e.g. to do a  clean -> `mvn flyway:clean -Plocal`)

**Reference**

Issues #543

## Summary by CodeRabbit

- **New Features**
	- Integrated Flyway for database migrations, enhancing schema management capabilities.
	- Added a new database table `the_entity` with defined attributes.
	- Introduced test data setup for the `the_entity` table.

- **Bug Fixes**
	- Removed outdated Hibernate configuration settings for improved clarity.

- **Documentation**
	- Added configuration details for Flyway migrations in both `application.yml` and `application-local.yml`.


refarch-backend/src/main/resources/db/migration/V001__TheEntity_schema.sql
```diff
@@ -0,0 +1,5 @@
+create table the_entity (
+    text_attribute varchar(8) not null,
+    id uuid not null,
+    primary key (id)
+);
```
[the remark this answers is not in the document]
eidottermihi: not the scope of this pr.

refarch-backend/src/main/resources/application-local.yml
```diff
@@ -3,12 +3,6 @@ server:
 spring:
   # Spring JPA
   jpa:
-    hibernate:
-      # always drop and create the db should be the best
-      # configuration for local (development) mode. this
-      # is also the default, that spring offers by convention.
-      # but here explicite:
-      ddl-auto: create-drop
```
devtobi, line 11: I'd suggest changing this to `validate` instead of completely removing it.

refarch-backend/src/main/resources/db/migration/V001__TheEntity_schema.sql
devtobi, line 1: We could separate the sql scripts into folders like `ddl` and `dml`. Also we might use flyway later on to prefill the database with testdata, so a `test` folder would be appropriate as well.
For now I think we should move this folder into a `ddl` subdirectory.

refarch-backend/pom.xml
devtobi, line 1: Would it be somehow possible to extract the database credentials from the `application,yml` file? Or is it possible to at set the DB credentials for Flyway in the `application.yml` files itself instead of hardcoding them into the `pom.xml`?

devtobi (changes requested):

refarch-backend/src/main/resources/application-local.yml
eidottermihi: done

refarch-backend/src/main/resources/application-local.yml
```diff
@@ -3,6 +3,9 @@ server:
 spring:
   # Spring JPA
   jpa:
+    hibernate:
+      # schema is managed by Flyway, only validate
+      ddl-auto: validate
```
devtobi, line 8: I think we need to move this `ddl-auto: validate` to the `application.yml` because we also need that when running in production, right?

devtobi (changes requested):

refarch-backend/src/main/resources/db/migration/V001__TheEntity_schema.sql
eidottermihi: I added two subfolders: `schema` and `testdata`.

`testdata` is only included in `flyway.locations` with `local` profile ATM.

I didn't go for your proposed `ddl` / `dml` structure as a single migration often can contain both, e.g. when adding a non-nullable column to a table:

- add a nullable column first: -> `ddl`
- set values for new column on existing data (UPDATE ... SET ...): `dml`
- make the colum non null: -> `ddl`

Or would you suggest that such a migration should be 3 scripts? e.g.

- `ddl/V001__add_nullable_column.sql`
- `dml/V002__set_default_values.sql`
- `ddl/V003__make_column_not_nullable.sql`

refarch-backend/pom.xml
eidottermihi: I don't think so because the Maven Flyway plugin (`pom.xml`) does not know anything about the Spring Boot Flyway integration (`application.yaml`).

The Maven plugin is also just a convenient way to execute Flyway against a specific db. I could remove the hardcoded user/password for the local stack postgres db from there, so that the user has to specify it as Maven property on execution?

`mvn flyway:clean -Plocal -Dflyway.user=<user> -Dflyway.password=<pass>`

devtobi: Nah I don't like the manual approach, then we should better keep the credentials inside `pom.xml`, especially as all other projects will have the same credentials for the database by default.

refarch-backend/src/main/resources/db/migration/V001__TheEntity_schema.sql
devtobi: I like your approach, lets keep it the `schema/testdata` way for now.

refarch-backend/src/main/resources/application.yml
```diff
@@ -1,6 +1,9 @@
 spring:
   application:
     name: @project.artifactId@
+  flyway:
+    locations:
+    - classpath:db/migration/schema
```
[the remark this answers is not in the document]
devtobi: @eidottermihi This looks interesting and useful, did you take a look at this?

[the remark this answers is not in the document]
eidottermihi: da hat der coderabbit etwas fantasiert, denn

clean-disbled: default ist bereits true
validate-on-migrate: default ist bereits true

zu baseline-on-migrate (https://documentation.red-gate.com/fd/baseline-on-migrate-224919695.html):

> This is useful for initial Flyway production deployments on projects with an existing DB.
>
> Be careful when enabling this as it removes the safety net that ensures Flyway does not migrate the wrong database in case of a configuration mistake!

Hat also nur entfernt was mit "Allows Flyway to work with existing databases" zu tun und macht in unserem Use Case keinen Sinn.

Ich würde also alles so lassen wie es ist: 
3dvurz

refarch-backend/pom.xml
DanielOber: You dont need the credentials for the DB inside the POM. Flyway takes the datasource, which is defined inside your application-*.yml files. 

If you run it in an environment, then you just have to add the environment variables to it and it will work.

In conclusion, please remove the configuration part from the POM.

devtobi: As far as I understood, the pom configuration is there to clear the database when the app itself it not running. I think this is still a useful feature for development as it "replicates" the "create-drop" functionality we had before (but running manually when desired).
This is way easier than connecting to the DB and clearing manually, or removing the docker volume.

devtobi: Just another IDE. I personally like running maven plugins via the IDE integration (e.g. in IntelliJ the tab on the right side of the IDE). This is often quicker than running via console directly. However because we require the profile `local` here, it cannot be run as easily with the IDE integration.
@eidottermihi Could we remove the profile binding as the plugin itself is only run manually anyways?

eidottermihi: done, now just the plugin is registered so that a manual execution is possible.

The rest (configuring url, user, password etc) now has to be done by the developer (e.g. by using IDE integration for maven, or configuring a user specific profile in users settings.xml).

devtobi: Why not copy over the credentials to the normal build configuration (combining both ideas). This way default credentials for database are already set, and no specific profile for execution is needed.

eidottermihi: I like to use the Maven plugin to also execute Flyway actions against test or even prod db. Defaulting the plugin settings to the local stack then feels strange, because then I would have to override most of the configs.

That was also why I initially moved the "local" stack configuration to a seperate Maven profile  😆

DanielOber: > Flyway actions against test or even prod db

That's why I prefer to disable that option entirely. I want to ensure it isn't used manually in any non-local environments. However, the current setup works for me. @devtobi, if you're satisfied with the solution as it stands, feel free to approve it.

devtobi: I don't like the fact that you now have the manual effort to provide db credentials yourself (e.g. setting them in the command or maven configuration). 
I still like the approach with the credentials in the pom more, as it removes the config on the developer side. If we now provide that configuration by default (no Maven profile) it will work for local development very easy. 
If you want to use it for other environments (which I discourage), you could still override the credentials in the command. @eidottermihi So can we combine both approaches as suggested by me few comments above?

eidottermihi: i restored url/user/password configuration, hopefully now its acceptable for every use case ;)

refarch-backend/pom.xml
```diff
@@ -439,6 +448,21 @@
...
                 </executions>
             </plugin>
+            <plugin>
+                <!-- to easily reset/clean local stack database: "mvn flyway:clean flyway:migrate -Dflyway.cleanDisabled=false" -->
+                <groupId>org.flywaydb</groupId>
+                <artifactId>flyway-maven-plugin</artifactId>
+                <configuration>
+                    <url>jdbc:postgresql://localhost:5432/postgres</url>
+                    <user>admin</user>
+                    <password>admin</password>
```
devtobi, line 458: It looks like the plugin even supports specific maven properties out of the box so we could move those parts up to the properties section to make configuration even easier when different environment is desired. See https://documentation.red-gate.com/fd/maven-goal-184127408.html#MavenGoal-Mavenproperties

refarch-backend/src/main/resources/db/migration/testdata/R__Testdata.sql
devtobi, line 1: Whats the reason for this file being named `R__`? :)

refarch-backend/pom.xml
eidottermihi: done

refarch-backend/pom.xml
```diff
@@ -439,6 +448,21 @@
...
+                    <locations>
+                        <location>classpath:db/migration/schema</location>
+                        <location>classpath:db/migration/testdata</location>
+                    </locations>
+                </configuration>
```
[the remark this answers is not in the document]
eidottermihi: clean is potentially very dangerous, so its better to leave it as is

refarch-backend/pom.xml
```diff
@@ -439,6 +454,19 @@
...
                 </executions>
             </plugin>
+            <plugin>
+                <!-- Flyway Maven plugin for local database management
+                     Usage:
+                     - Clean database: mvn flyway:clean
+                     - Apply migrations: mvn flyway:migrate
+                     - Reset and migrate: mvn flyway:clean flyway:migrate
+                     
+                     Note: Use -Dflyway.cleanDisabled=false when running clean command
```
devtobi, line 464: Why not set cleanDisabled to false by default? Is this too risky? I think when I execute the clean goal I know what I'm doing.

eidottermihi: I was using the flyway-maven-plugin to do other stuff (like baselining a prod db to reduce hundreds of migrations to 1) in the past. So therefore I think it's safer not to set this configuration, because accidentally running clean on prod can make a busy day ;)

devtobi: Ok fine for me, lets get this thing merged ;)

devtobi (approved):
LGTM

PR #633, opened 2024-12-10
:recycle: backend cleanup local sso conf

simonhir:
**Description**

Cleanup backend local sso configuration.
Use keycloak hostname `localhost` instead of `keycloak` so that users do not have to customize the hosts file.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new Single Sign-On (SSO) configuration for enhanced security and modularity.
- **Bug Fixes**
	- Updated security properties to improve integration with identity providers.
- **Chores**
	- Removed outdated Keycloak configuration to streamline settings.

### DIFF
--- a/refarch-backend/src/main/resources/application-local.yml
+++ b/refarch-backend/src/main/resources/application-local.yml
@@ -26,7 +26,7 @@ spring:
           audiences:
             - ${sso.client}
 
-# Define the local keycloak realm here
+# Define the local keycloak configuration here
 sso:
   url: http://localhost:8100
   realm: local_realm
--- a/refarch-backend/src/main/resources/application-local.yml
+++ b/refarch-backend/src/main/resources/application-local.yml
@@ -27,6 +27,7 @@ spring:
             - ${sso.client}
 
 # Define the local keycloak configuration here
+# Deduplication-only property
 sso:
   url: http://localhost:8100
   realm: local_realm
--- a/refarch-backend/src/main/resources/application-local.yml
+++ b/refarch-backend/src/main/resources/application-local.yml
@@ -22,16 +22,17 @@ spring:
     oauth2:
       resourceserver:
         jwt:
-          jwk-set-uri: ${keycloak.auth-server-url}/auth/realms/${realm}/protocol/openid-connect/certs
+          jwk-set-uri: ${sso.url}/auth/realms/${sso.realm}/protocol/openid-connect/certs
           audiences:
-            - local
+            - ${sso.client}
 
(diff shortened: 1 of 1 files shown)


devtobi:
@simonhir I tought the problem was already fixed with https://github.com/it-at-m/refarch-templates/pull/536?

refarch-backend/src/main/resources/application-local.yml
```diff
@@ -22,16 +22,17 @@ spring:
...
 
-# Define the local keycloak realm here
-realm: local_realm
-keycloak:
-  auth-server-url: http://keycloak:8100
+# Define the local keycloak configuration here
+sso:
+  url: http://localhost:8100
+  realm: local_realm
+  client: local
```
devtobi, line 33: I like that you grouped those attributes together. But should we place it under `refarch` key? I know those are not properties we use inside the code and are just placed here to avoid duplication but still.

devtobi:
> > @simonhir I tought the problem was already fixed with #536?
> 
> actually no, you can test it, if you try to access the backend from the frontend you will get an `unkown` host error. So it has to be either `http://localhost:8100` or `http://kubernetes.docker.internal`, both work.

Ah I see, so its not a Gateway <-> Backend thing, but a Backend <-> SSO thing, sorry I somehow misunderstood that. 

Yeah then I think `localhost` is the way to go, because the keycloak server is exposed locally via `8100` right? Or are there any advantages if we use `kubernetes.docker.internal`?

refarch-backend/src/main/resources/application-local.yml
DanielOber: i dont really see a benefit by doing that. I would prefer how it is currently done

DanielOber:
> > > @simonhir I tought the problem was already fixed with #536?
> > 
> > 
> > actually no, you can test it, if you try to access the backend from the frontend you will get an `unkown` host error. So it has to be either `http://localhost:8100` or `http://kubernetes.docker.internal`, both work.
> 
> Ah I see, so its not a Gateway <-> Backend thing, but a Backend <-> SSO thing, sorry I somehow misunderstood that.
> 
> Yeah then I think `localhost` is the way to go, because the keycloak server is exposed locally via `8100` right? Or are there any advantages if we use `kubernetes.docker.internal`?

dont really see any advantages or disadvantages of using `localhost` or `kubernetes.docker.internal`. For now, i would also go with `localhost`

devtobi (approved):
LGTM

DanielOber (approved):
LGTM

Issue #434, opened 2024-10-18
[Feature] Add frontend and webcomponent preview servers to gateway route configuration


devtobi:
### Relevant template

refarch-frontend, refarch-webcomponent

### Desired solution

Using vite preview (which starts a fully fledged webserver that is able to replicate nginx in production) we can deploy the frontend application and webcomponent application behind the gateway using additional routes.
We need to make sure both preview servers use different ports (configurable via vite config).

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #635, opened 2024-12-11
Maint/434 feature add frontend and webcomponent preview servers to gateway route configuration

devtobi:
**Description**
- Updated configurations to use local frontend Vite dev server behind API gateway

I explictly did not include it for webcomponents as it does not make sense for local development to use the webcomponent dev server behind the gateway. (only loader files are interesting for webcomponents which are only created on build)

**Reference**
Issues #434

## Summary by CodeRabbit

- **New Features**
	- Introduced a new route configuration in the `refarch-gateway` service for improved traffic handling.
	- Enhanced accessibility of the Vite server for development across different devices.

- **Bug Fixes**
	- Standardized healthcheck configurations across multiple services for better monitoring and reliability.

- **Chores**
	- Removed the preview script from the build process.

### DIFF
--- a/refarch-frontend/package.json
+++ b/refarch-frontend/package.json
@@ -10,7 +10,6 @@
     "dev": "vite",
     "test": "vitest run",
     "build": "vue-tsc --build --noCheck && vite build",
-    "preview": "vite preview",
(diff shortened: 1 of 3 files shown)


simonhir (approved):
LGTM

Issue #652, opened 2024-12-16
[Documentation] Create Get started


DanielOber:
### Relevant documentation

refarch-frontend, refarch-backend, refarch-eai, refarch-webcomponent

### Desired solution

Create a get started for node and maven projects. Describe the necessary steps and requirements to use the templates

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #654, opened 2024-12-17
doc/652-documentation-create-getting-started

DanielOber:
**Description**

- added getting started guide for all templates

**Reference**

Issues #652 

## Summary by CodeRabbit

- **New Features**
	- Introduced a comprehensive guide for users in the new `getting-started.md` document.
	- Updated the Vitepress configuration to include a new "Getting Started" entry in the sidebar.

- **Bug Fixes**
	- Improved markdownlint configuration by adding a rule to prevent duplicate headings.

- **Documentation**
	- Added detailed instructions for setting up a code repository and using templates in the new guide.


docs/getting-started.md
```diff
@@ -0,0 +1,100 @@
...
+**PGAdmin**
+- **Local Instance**: Run a pre-configured instance of PgAdmin that connects to your local PostgreSQL database.
+
+**API Gateway**
+- **Local Instance**: Launch the ITM API-gateway in your local environment.
+
+**AppSwitcher**
+- **Repository**:https://github.com/it-at-m/appswitcher-server
+- **Configuration**: Configure the AppSwitcher within the `stack/appswitcher-server` folder.
+
```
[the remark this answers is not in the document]
devtobi: I like the idea of adding link to external documentation and the running port in our development stack. I encourage we do this for every container image in the stack. @DanielOber What do you think about this?

docs/getting-started.md
```diff
@@ -0,0 +1,100 @@
...
+- Configured to operate without security measures for development
+  or testing purposes.
+
+
+## Frontend and Web Components
+
+### Prerequisites
+
+- Node.js version 20 to 22
+- Docker (AppSwitcher)
```
devtobi, line 112: Would not mention Docker here as its already defined in "Getting started".

docs/getting-started.md
```diff
@@ -0,0 +1,100 @@
...
+as a general setup and requirements applicable to all templates.
+
+## General
+
+To use the refarch-templates, you need to have Docker installed on your
+machine. Inside the `stack` folder, you will find a `docker-compose.yml`
+file that will spin up everything needed for local development. 
+
+When developing locally, the stack should always be running. 
+### Download Templates
```
devtobi, line 18: I'd add documentation for the alternative way. Instead of cloning it, users could also use the template directly via GitHub, as we provided the repository as a "GitHub template repository"

docs/getting-started.md
```diff
@@ -0,0 +1,100 @@
+# Getting Started
+
+Welcome to the Getting Started guide for the refarch-templates. This guide
+is divided into sections for Frontend/Web Component and Backend/EAI, as well
+as a general setup and requirements applicable to all templates.
```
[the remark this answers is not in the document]
devtobi: I'd add a small sentence about the purpose of the templates like: "Streamline development of web applications according to the reference architecture of it@M" and include a link. @DanielOber What do you think?

docs/getting-started.md
```diff
@@ -0,0 +1,100 @@
...
+file that will spin up everything needed for local development. 
+
+When developing locally, the stack should always be running. 
+### Download Templates
+
+You can download the templates or clone them directly from the GitHub
+repository.
+```
+git clone https://github.com/it-at-m/refarch-templates.git
+```
```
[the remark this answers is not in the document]
devtobi: I agree with CodeRabbit to include the language inside the code block, as this improves syntax highlighting.

docs/getting-started.md
```diff
@@ -0,0 +1,100 @@
...
+## General
+
+To use the refarch-templates, you need to have Docker installed on your
+machine. Inside the `stack` folder, you will find a `docker-compose.yml`
+file that will spin up everything needed for local development. 
+
+When developing locally, the stack should always be running. 
+### Download Templates
+
+You can download the templates or clone them directly from the GitHub
```
devtobi, line 20: I don't think we should recommend cloning here. Cloning is only relevant when contributing to the templates, not when using them.

docs/getting-started.md
```diff
@@ -0,0 +1,119 @@
...
+
+- **Local Instance**: Set up and run a PostgreSQL instance locally.
+
+### PGAdmin
+
+- **Local Instance**: Run a pre-configured instance of PgAdmin that connects to
+your local PostgreSQL database.
+
+### API Gateway
+
```
devtobi, line 58: Do we need such specific information? Or is Maven 3.9 or even Maven 3 enough information?

docs/getting-started.md
```diff
@@ -0,0 +1,119 @@
...
+- **Local Instance**: Launch the ITM API-gateway in your local environment.
```
devtobi, line 59: Remove as already mentioned in getting started.

docs/getting-started.md
```diff
@@ -0,0 +1,119 @@
...
+```
+
+### Profiles
+
+By default, the template supports two profiles:
+
+#### local
+
+- Contains the necessary configurations for running the application
+  locally, including settings for ports, SSO, and the database.
```
devtobi, line 100: Why is this needed?

docs/getting-started.md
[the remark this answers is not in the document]
DanielOber: added one, but can change it

docs/getting-started.md
DanielOber: that kinda depends, is that the way people should use it? If so, can we chance the interel process, so that a new project that is based on the refarch will get automatically get the template? Should discuss this internally first.

docs/getting-started.md
DanielOber: I let 2 People outside of the team, review it, and one of the points was, that it would be good to get a reminder, that something from the stack will be needed, thats why i would leave it in.

docs/getting-started.md
DanielOber: but it our standard maven, so we now that it works with it. Does it work with maven 3.8? maven 3.9.1? 3.7? i didnt check and i dont think that there is any reason to check it.

docs/getting-started.md
DanielOber: also got it from the review, that it would be nice to see why the profiles are needed and have a small explanation

docs/getting-started.md
DanielOber: cloning or downloading doesnt matter, but cloning is more convenient, because you dont get a zip and the one command is also better then looking how to get the zip.

docs/getting-started.md
```diff
@@ -0,0 +1,166 @@
...
+
+1. Rename the folder you copied (e.g. from `refarch-backend` to `myapp-backend`).
+2. Change the package name from `de.muenchen.refarch` to
+   `de.muenchen.YourPackageName` (It's recommended to use an IDE for this
+   task, as it will automatically update the imports; otherwise, you will
+   need to do this manually).
+3. Inside the `pom.xml`, update the `artifactId`, `groupId`, and `name`
+   fields.
+4. Install required dependencies by executing the command `mvn install`.
+
```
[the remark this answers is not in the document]
devtobi: @DanielOber What do you think about this? Should we add a small section on how users can verify if their setup worked?

docs/getting-started.md
```diff
@@ -0,0 +1,158 @@
...
+The purpose of these templates is to provide developers with an out-of-the-box solution for
+building full-stack applications based on the reference architecture (RefArch) of it@M.
+For more information on the RefArch itself please visit the [RefArch documentation](https://refarch.oss.muenchen.de/)
+
+**Important**: To use the `refarch-templates` you need to have [Docker](https://www.docker.com/) installed on your system.
+
+## Getting the templates
+
+There are two different ways on how to get started with using the templates. Which variant to use depends on whether you
```
simonhir, line 15:
```suggestion
There are two different ways on how to get started with the templates. Which variant to use depends on whether you
```

docs/getting-started.md
```diff
@@ -0,0 +1,158 @@
...
+### Configure templates
+
+After getting the templates, you will need to make a few adjustments:
+
+1. Rename the folder you copied (e.g. from `refarch-backend` to `myapp-backend`).
+2. Change the package name from `de.muenchen.refarch` to
+   `de.muenchen.YourPackageName` (It's recommended to use an IDE for this
+   task, as it will automatically update the imports; otherwise, you will
+   need to do this manually).
+3. Inside the `pom.xml`, update the `artifactId`, `groupId`, and `name`
```
simonhir, line 121:
```suggestion
3. Inside the `pom.xml`, update the `artifactId`, `groupId`, `name` and `description`
```

docs/getting-started.md
```diff
@@ -0,0 +1,158 @@
...
+   fields.
+4. Install required dependencies by executing the command `mvn install`.
```
simonhir, line 123:
```suggestion
4. Install required dependencies by executing the command `mvn install`.
```

Probably rename the Main Class? Think most projects do this and makes running multiple clearer

docs/getting-started.md
```diff
@@ -0,0 +1,158 @@
...
+
+Congratulations, the Backend/EAI is now ready to use.
+
+### Profiles
+
+By default, the template supports two Maven profiles:
+
+- `local`: Contains the necessary configurations for running the application
+  locally, including settings for ports, SSO, and the database
+- `no-security`: Configured to operate without security measures for development
```
simonhir, line 133: What's about `json-logging`?

docs/getting-started.md
```diff
@@ -0,0 +1,158 @@
...
+  or testing purposes.
+
+## Frontend & Web Components
+
+### Prerequisites
+
+- Node.js version 20 to 22
```
simonhir, line 140: Why both versions and not just 22? This requires us for always supporting both

docs/getting-started.md
```diff
@@ -0,0 +1,158 @@
...
+3. Generate a new `package-lock.json` file by executing the command `npm install`.
+
+Congratulations, the Frontend/WebComponent is now ready to use.
+
+## Next steps
+
+If you finished the template configurations for all required components your project needs
+you have now successfully set up your project.
+
+To help you with software developing and project organisation, please check out **TODO** and **TODO**.
```
simonhir, line 158: Why don't you add dead links as i did? Think that's more useful than that or just write where it should link in the feature

devtobi: Actually we don't know how many links there will be yet so I just set it to TODO. 😄

docs/getting-started.md
devtobi: `json-logging` is not relevant for local development, so I wouldn't include it here.

docs/getting-started.md
devtobi: Will discuss this next year.

docs/getting-started.md
```diff
@@ -0,0 +1,158 @@
...
+- Maven 3.9.7
+- Docker (for Postgres and Keycloak)
+
+### Configure templates
+
+After getting the templates, you will need to make a few adjustments:
+
+1. Rename the folder you copied (e.g. from `refarch-backend` to `myapp-backend`).
+2. Change the package name from `de.muenchen.refarch` to
+   `de.muenchen.YourPackageName` (It's recommended to use an IDE for this
```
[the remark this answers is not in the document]
devtobi: Will discuss this next year.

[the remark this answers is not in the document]
devtobi: https://git.muenchen.de/ccse/ospo/-/issues/532

simonhir (approved):

devtobi (approved):
LGTM

Issue #777, opened 2025-02-11
[Bug] maven build produces incompatible named jar


langehm:
### Bug description

The maven install produces a jar named `refarch-backend-X.X.X.jar` which is not compatible with the Dockerfile. The requested name of the jar in the Dockerfile is `application.jar`.

Therefore the execution of the build container fails ... no manifest, no jar found.

### Expected behaviour

The maven install process should produces a jar called `application.jar` to always be compatible with the Dockerfile.

### Steps for reproduction

1. clone latest project
2. cd refarch-backend
3. mvn install
4. cd target
5. ...

### Affected version

latest

### Affected template

refarch-backend

### Relevant log output (optional)

```shell
INFO running in /deployments
no main manifest attribute, in /deployments/application.jar
```

### Screenshots (optional)

Target folder:
![Image](https://github.com/user-attachments/assets/78a49636-52a2-4c25-848c-ca719a34c5c6)

Dockerfile:

![Image](https://github.com/user-attachments/assets/c1e70534-3872-4706-86e1-dde0ad8410fc)

### No duplicate

- [x] I confirm that this issue is not a duplicate

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

hupling:
I don't understand why @tobiasholler doesn't have the problem with the https://github.com/it-at-m/captchaservice Unfortunately the repo is private and I can't compare.

hupling:
The problem is that not the refarch-backend jar will be copied. The findebugs.jar was copied.

devtobi:
That is a good point. So a good solution would be to improve the copying mechanism. @simonhir Didn't you already have a solution for this in your RefArch-based project?

langehm:
> Actually it should work because the Dockerfile copies over any jar file and names it "application.jar" when copying. This allows us to stay independent of the Maven build process itself.

I see it now - misread the Dockerfile. But the issue still persists as @hupling mentioned.

> So a good solution would be to improve the copying mechanism

... probably yeah

langehm:
A cheap solution could be sth anlong this line: 
```
# Copy runnable jar to deployments
COPY target/*-*.*.*.jar /deployments/application.jar
```

simonhir:
I explicitly changed the copy command to only match the correct jar:

```
COPY target/dispatch-service-*.jar /deployments/application.jar
```

devtobi:
[...]

PR #778, opened 2025-02-11
:bug: changed the name of the build jar to application.jar

langehm:
**Description**

- changed the build jar name to be compatible with the Dockerfile

**Reference**

Issues #777

## Summary by CodeRabbit

- **Chores**
	- Updated the build configuration so that the packaged artifact is consistently named "application". This change ensures uniformity across builds without affecting existing functionality.

### DIFF
--- a/refarch-backend/pom.xml
+++ b/refarch-backend/pom.xml
@@ -310,6 +310,8 @@
     </scm>
 
     <build>
+        <!-- Define the name of the jar -->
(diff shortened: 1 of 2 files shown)


devtobi:
I don't think this is a good solution because we want to be able to release the jar files later on as well (and not only the docker images). So it's a must that the jar files retain the application name.

langehm:
Probably not needed anymore

Issue #184, opened 2024-08-09
[Maintenance] Add keycloak roles to stack


devtobi:
### Problem description (optional)

Currently the Permissions defined in AuthoritiesEnum of the backend used to secure the endpoints are not available in the docker stack. This leads to the backend not working as intended by default.

### Desired solution

- Updated the PreAuthorize annotations to check for role instead of authority
- Add roles to keycloak client (writer, reader role)
- Add users to keycloak client
    - none:none -> no role
    - writer:writer -> writer role
    - reader:reader -> reader role
- Additionally specific mappers might be necessary
- Update documentation

### Considered alternatives (optional)

---

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

devtobi:
Further discussion about roles vs authorities in keycloak required. [keycloakmigratio ](https://mayope.github.io/keycloakmigration/) does currently not support the configuration of client authorization settings.

devtobi:
See https://github.com/it-at-m/refarch-templates/discussions/191 for further discussion. This issue will be blocked until then.

devtobi:
Ticket is no longer blocked, as a internal decision was made to provide role based authorization by default. This means only roles need to be added to the stack.

PR #792, opened 2025-02-14
Maint/184 maintenance add keycloak roles and permissions to stack

devtobi:
**Description**
- Added more users
- Added more roles
- Split migration scripts into definition and attachment of roles
- Removed user_name custom attribute and replaced with preferred_username
- Change Authorization enum to use `hasRole` instead of `hasAuthority`

**WIP**:
Currently we have the problem, that the user info endpoint cannot be fetched by the backend using the token it received from the API gateway. However when calling the userinfo point directly from the gateway (via `localhost:8083/api/sso/userinfo`) everything works fine. Copying the token into postman and manually requesting the userinfo from the keycloak server also works fine.

**Reference**
Issues #184

## Summary by CodeRabbit

- **Documentation**
  - Updated local development instructions to clarify Docker setup and host file configuration.

- **New Features**
  - Enhanced access control by transitioning to more granular, role-based permissions for various operations.
  - Improved identity management with an updated SSO endpoint and refined user/role assignments.
  - Added new roles (`reader`, `writer`) with descriptions for access levels in the application.
  - Introduced a new user management configuration for adding multiple users with specific attributes.


simonhir:
Probably that's the problem
grafik

refarch-backend/src/main/resources/application-local.yml
```diff
@@ -27,7 +27,7 @@ spring:
 # Define the local keycloak configuration here
 # Deduplication-only property
 sso:
-  url: http://localhost:8100
+  url: http://kubernetes.docker.internal:8100
```
simonhir, line 30: We definitely should add this to the docs

devtobi: This should happen when Docker is installed automatically on most systems right? However a little note should still be handy. ;)

refarch-backend/src/main/java/de/muenchen/refarch/security/Authorities.java
```diff
@@ -1,20 +1,22 @@
...
 
-import org.springframework.data.repository.PagingAndSortingRepository;
 import org.springframework.security.access.prepost.PreAuthorize;
 
 /**
  * Each possible authority in this project is represented by an enum.
- * The enums are used within the {@link PagingAndSortingRepository}
+ * The enums are used within the {@link org.springframework.stereotype.Controller} or
+ * {@link org.springframework.stereotype.Service} classes.
  * in the annotation e.g. {@link PreAuthorize}.
```
simonhir, line 9:
```suggestion
 * {@link org.springframework.stereotype.Service} classes in the method security annotations (e.g. {@link PreAuthorize}).
```

simonhir (approved):
LGTM

simonhir (approved):
LGTM

Issue #78, opened 2024-07-23
Change internal links to useful Public links


DanielOber:
**Is your feature request related to a problem? Please describe.**
There are some links, that reference internal documentation and can not be used by everyone

**Describe the solution you'd like**
change the links to public documentations

PR #829, opened 2025-02-28
change internal links to useful external links

DanielOber:
# Pull Request

[documentation-link]: https://refarch-templates.oss.muenchen.de/document.html#writing-the-documentation
[code-quality-link]: https://refarch-templates.oss.muenchen.de/develop.html#code-quality

## Changes

- changed the links on the landing page

@devtobi there is still one internal link in the [user-client](https://github.com/it-at-m/refarch-templates/blob/main/refarch-frontend/src/api/user-client.ts#L14). I think for now we should leave it, because we still need to talk about some configs. What do you think? 

## Reference

Issue: #78 

## Summary by CodeRabbit

- **Documentation**
  - Streamlined the documentation section on the "Get Started" page.
  - Updated primary hyperlinks with new URLs and display texts.
  - Removed obsolete links to ensure users access the most current resources.


devtobi:
I think we should remove the internal link in the user client as well, because long term we don't want to have any LHM specific scopes inside this file, do we?

DanielOber:
> I think we should remove the internal link in the user client as well, because long term we don't want to have any LHM specific scopes inside this file, do we?

thats right, but perhaps there will be changed in the future to the keycloak mappers that we currently got. So later on, we maybe can make or link better documentation. For now, the internal link does not hurt imo, but if you think we should remove it i will delete it for now.

devtobi:
> > I think we should remove the internal link in the user client as well, because long term we don't want to have any LHM specific scopes inside this file, do we?
> 
> thats right, but perhaps there will be changed in the future to the keycloak mappers that we currently got. So later on, we maybe can make or link better documentation. For now, the internal link does not hurt imo, but if you think we should remove it i will delete it for now.

I just wanna make sure we don't forget to update the documentation in the future. However its fine for me like this as well. Do as you please. ;)

devtobi (approved):
LGTM

Issue #258, opened 2024-08-29
[Maintenance] Keycloak custom mappers


devtobi:
### Problem description (optional)

Currently our user object in `refarch/frontend/types/User.ts` contains lots of metadata which come from our internal SSO provider. However we should modify our local keycloak stack to provide this data as well (dummy data).
Missing information in the local stack will lead to errors in the frontend, because User object cannot be constructed safely.

### Desired solution

Modify the local keycloak stack to reproduce our company internal user information.

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

simonhir:
The check of the `aud` claim is also relevant here as we currently use a custom mapper/scope for that.
See https://github.com/it-at-m/refarch-templates/blob/main/refarch-backend/src/main/resources/application-local.yml#L26-L27 and  https://github.com/it-at-m/refarch-templates/blob/main/stack/keycloak/migration/06_local-audience.yml

simonhir:
Will remove all custom mappers with https://github.com/it-at-m/refarch-templates/pull/1538. the lhm-core scopes will be readded (optional per project) via #1387

PR #1527, opened 2026-04-16
Maintenance: Replace classes with interfaces

### LINKED ISSUE #1524: [Maintenance] Replace classes with interfaces
### Relevant template

frontend, webcomponent

### Problem description (optional)

The templates contain some classes that act as a simple data container.
This is unnecessary overhead as classes have a higher complexity than plain JS objects.

### Desired solution

Refactor classes to TypeScript `interfaces` to simplify the code and improve consistency.

### Considered alternatives (optional)

_No response_

### Additional context (optional)

_No response_

### No duplicate

- [x] I confirm that this issue is not a duplicate

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

devtobi:
# Pull Request

[code-quality-link]: https://refarch.oss.muenchen.de/templates/develop#code-quality
[refarch-create-documentation-issue-link]: https://github.com/it-at-m/refarch/issues/new?template=4-documentation-change.yml

## Changes

- Replaced `HealthState` class with an interface
- Replaced `Ad2ImageAvatarClient` class with an exported method (configured via the constant through env variable)
- Replaced `User` class with an interface
- Renamed `User` to `UserInfo` throughout the whole source files to make more clear its related userinfo endpoint
- Renamed `Info` to `AppInfo` throughout the whole source files (to align with UserInfo), Info was a bit to generic.

## Reference

Issue: #1524

## Checklist

**Note**: If some checklist items are not relevant for your PR, just remove them.

### General

- [x] Met all acceptance criteria of the issue
- [x] Added meaningful PR title and list of changes in the description

### Code

- [x] Wrote code and comments in English

## Summary by CodeRabbit

* **Refactor**
  * Switched the app to a unified UserInfo model with a new user-info state store and updated initialization flow.
  * Consolidated avatar URL generation into a single utility and updated components to use it.
  * Standardized app info and health typings and cleaned up internal import paths.

* **Bug Fixes**
  * Improved default/fallback behavior and streamlined user-info-related error messaging.


frontend/src/api/ad2imageavatar-client.ts
```diff
@@ -0,0 +1,25 @@
...
+ */
+export function getAvatarHref(
+  username: string,
+  mode = "fallbackGeneric",
+  size = "64"
+): string {
+  if (!AD2IMAGE_URL) {
+    throw new Error(
+      "VITE_AD2IMAGE_URL was not configured. Please use the .env files."
+    );
```
langehm: ON second thought - throw will result in a hard crash. Could this be soften up by using alternative values?

frontend/src/api/ad2imageavatar-client.ts
```diff
@@ -0,0 +1,25 @@
...
+  }
+  const url = new URL("/avatar", AD2IMAGE_URL);
```
langehm, line 15: If this is always `/avatar` - maybe this could be moved to a more permanent place.

frontend/src/api/userinfo-client.ts
```diff
@@ -13,35 +15,20 @@ import User from "@/types/User";
...
  */
-export function getUser(): Promise<User> {
+export function getUserInfo(): Promise<UserInfo> {
   return fetch("api/sso/userinfo", getConfig())
     .catch(defaultCatchHandler)
     .then((response) => {
       defaultResponseHandler(
         response,
-        "Beim Laden des Users ist ein Fehler aufgetreten."
+        "Beim Laden der UserInfo ist ein Fehler aufgetreten."
```
langehm, line 24: mhh we cannot use i18n here, right?

frontend/src/stores/userinfo.ts
langehm, line 1: Mhh this file was originally moved - now this file shows as being newly created.

frontend/src/types/UserInfo.ts
```diff
@@ -0,0 +1,63 @@
...
+  displayName: "Local Development User",
+  surname: "User",
+  telephoneNumber: "+1234567890",
+  email: "local@development.com",
+  username: "Local Development User",
+  givenname: "Local",
+  department: "Local Department",
+  lhmObjectID: "LOCAL_ID",
+
+  // LHM_Extended
```
langehm, line 34: We should not promote this scope anymore as it got deprecated

frontend/src/types/UserInfo.ts
```diff
@@ -0,0 +1,63 @@
...
+  preferred_username: "",
+  memberof: [],
+  user_roles: [
+    // todo add authorities
+  ],
+  authorities: [
+    // todo add user roles
+  ],
```
langehm, line 42: Same as above with this specific field - maybe add a `depreacted` note?

frontend/src/types/UserInfo.ts
```diff
@@ -0,0 +1,63 @@
...
+  displayName: "",
+  surname: "",
+  telephoneNumber: "",
+  email: "",
+  username: "",
+  givenname: "",
+  department: "",
+  lhmObjectID: "",
+
+  // LHM_Extended
```
langehm, line 58: same here

langehm (changes requested):
Looks good otherwise - just some questions

frontend/src/api/ad2imageavatar-client.ts
devtobi: What do you mean with hard crash? In the end its just a error message that gets logged to the console.

frontend/src/api/ad2imageavatar-client.ts
devtobi: I could create a constant variable like `AVATAR_PATH` inside the client file. However its not reused anywhere. I definitely wouldn't move it to `constants.ts` or something.

frontend/src/api/userinfo-client.ts
devtobi: i18n in parts other than the UI is a whole different topic. This would affect the whole app. But we should investigate it. Please open an issue for that topic.

frontend/src/stores/userinfo.ts
devtobi: Don't know why. Had those problems in the past in other PRs from other persons as well. Would ignore it for now.

frontend/src/types/UserInfo.ts
devtobi: Overhaul of the UserInfo class as a whole is related to https://github.com/it-at-m/refarch-templates/issues/258 and not part of this PR.

frontend/src/api/ad2imageavatar-client.ts
langehm: Is it in vue? This function will be stopped as far as i remember. By crashing a component could be hindered to render.

frontend/src/types/UserInfo.ts
devtobi: Part of #258

frontend/src/types/UserInfo.ts
devtobi: Part of #258

frontend/src/api/userinfo-client.ts
langehm: No you ...
unocardds2_2048x2048

frontend/src/api/userinfo-client.ts
simonhir, line 1: Why did you rename that?

devtobi: To make it more clear, that its related to the UserInfo endpoint. :)

frontend/src/api/userinfo-client.ts
devtobi: Will be done in https://github.com/it-at-m/refarch-templates/issues/1529

frontend/src/api/userinfo-client.ts
simonhir: I liked the previous more. I think that's a mixing of concerns. In the application it should be totally irrelevant where the info came from. Otherwise you would need to name something like `TestService` `TestDbService`

devtobi: @langehm What do you think about this?

langehm: mhh every other api-call in the future will be generated by openapi. From my perspective these could specific to which "endpoint" they use internally.

simonhir: @langehm under Java the Spring classes follow the description not the underlying API, think that would be the same for JS and then they should also look like getUser and not getUserInfo or?

langehm: I understand your point - from my pov i dont fetch a user - i get some technical informations about the user.

devtobi: Discussed this and will change to `UserInfo` for now. If we see that its cumbersome to use, we are open to change that back in the future.

frontend/src/api/ad2imageavatar-client.ts
devtobi: Requires internal discussion, thus will be handled in a later Issue/PR

frontend/src/types/UserInfo.ts
```diff
@@ -0,0 +1,63 @@
...
+  lhmObjectID: string;
+
+  // LHM_Extended
+  preferred_username: string;
+  memberof: string[];
+  user_roles: string[];
+  authorities: string[];
+}
+
+export const USERINFO_LOCAL_DEVELOPMENT: UserInfo = {
```
langehm, line 21: Should we mix types with constants?
Same for `USERINFO_EMPTY`

frontend/src/App.vue
```diff
@@ -16,35 +16,37 @@
...
+import { useUserInfoStore } from "@/stores/userinfo";
+import { USERINFO_LOCAL_DEVELOPMENT } from "@/types/UserInfo";
 
-const userStore = useUserStore();
+const userInfoStore = useUserInfoStore();
 const [isNavigationShown, toggleNavigation] = useToggle();
 
 onMounted(() => {
-  loadUser();
+  loadUserInfo();
```
langehm, line 35: Loading the user should be done at an earlier stage than here. But this should be addressed in a separate issue - can you create one?

For context: The current implementation will load the user after e.g. the vue-router which could create some issues. But i think we also have a plan for reworking the rolemamagement... could be addressed there

langehm (changes requested):

frontend/src/App.vue
devtobi: Instead of creating an issue, let's talk about this in our next meeting first. I added it to the agenda.

langehm (approved):

Issue #1529, opened 2026-04-17
[Maintenance] i18n in parts besides Vue components


devtobi:
### Relevant template

frontend

### Problem description (optional)

Currently i18n is used inside Vue components using its composable.
To increase the coverage of i18n, we should extend the usage to script files as well (e,g. error logging, toast messages, ...).
There currently is no unified way to accomplish that.

### Desired solution

Extend i18n usage to other parts than Vue components.

### No duplicate

- [x] I confirm that this issue is not a duplicate

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

devtobi:
Translation file access outside of Vue components is possible via `i18n.global.t`, where i18n is the exported plugin defined in `plugins/i18n.ts`.

PR #1607, opened 2026-05-22
Feature: Add composable and navigation guards to secure UI and routes

### LINKED ISSUE #607: [Feature] Add navigation guard to secure routes via route metadata
### Relevant template

refarch-frontend

### Problem description (optional)

Currently there is no out-of-the-box mechanism to secure routing to unauthorized routes.

### Desired solution

Introduce a composable which checks a required role against the users current role.
This should be used inside a global route guard which inspects potential metadata attached to the routes to secure the route navigation.
We could add metadata like `roles` or `authorities` to the routes - metadata field should be typed and compatible to the new vue-router v5.

Secondly, this composable can be used to display or hide elements according to the required role.

### Considered alternatives (optional)

_No response_

### Additional context (optional)

_No response_

### No duplicate

- [X] I confirm that this issue is not a duplicate

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

devtobi:
# Pull Request

[code-quality-link]: https://refarch.oss.muenchen.de/templates/develop#code-quality
[refarch-create-documentation-issue-link]: https://github.com/it-at-m/refarch/issues/new?template=4-documentation-change.yml

## Changes

- Added composable for checking user roles against a set of required roles (e.g. to hide UI elements)
- Added custom route guard using route metadata to secure routing to a specific route
- Added example usages of the two new features

## Reference

Closes #607

## Checklist

**Note**: If some checklist items are not relevant for your PR, just remove them.

### General

- [x] Met all acceptance criteria of the issue
- [x] Added meaningful PR title and list of changes in the description
- [x] Opened documentation issue in [refarch repository](https://github.com/it-at-m/refarch/pull/890)

### Code

- [x] Wrote code and comments in English
- [x] Removed waste on branch (e.g. `console.log`), see [code quality tooling][code-quality-link]

#### Frontend / WebComponent

- [x] Tested changes using API gateway

## Summary by CodeRabbit

* **New Features**
  * Role-based access control for routes and navigation links.
  * Role-aware content on the index and "Get started" pages (shows different text for writers vs non-writers).
  * New role utilities to check user roles.

* **Chores**
  * Updated authentication client configuration for stricter scope handling.

* **Localization**
  * Added German strings for writer/non-writer messaging.

[Review Change Stack](https://app.coderabbit.ai/change-stack/it-at-m/refarch-templates/pull/1607?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


frontend/src/composables/useHasAnyRole.ts
```diff
@@ -0,0 +1,22 @@
...
+
+export default function useHasAnyRole(roles: Role | Role[]) {
+  const userInfoStore = useUserInfoStore();
+  return computed(() => hasAnyRole(roles, userInfoStore.currentRoles));
+}
+
+/**
+ * Helper function: Only used in route guard, do not use directly
+ * @param roles
+ * @param currentRoles
```
simonhir: Why add this without description?

frontend/src/composables/useHasAnyRole.ts
```diff
@@ -0,0 +1,22 @@
...
+
+import { useUserInfoStore } from "@/stores/userinfo";
+
+export default function useHasAnyRole(roles: Role | Role[]) {
+  const userInfoStore = useUserInfoStore();
+  return computed(() => hasAnyRole(roles, userInfoStore.currentRoles));
+}
+
+/**
+ * Helper function: Only used in route guard, do not use directly
```
simonhir:
```suggestion
 * Helper function: Only use directly in route guard. Use useHasAnyRole elsewhere.
```

frontend/src/plugins/router.ts
```diff
@@ -22,12 +23,25 @@ const router = createRouter({
...
+    !to.meta.hasAnyRole ||
+    hasAnyRole(to.meta.hasAnyRole, userInfoStore.currentRoles)
+  ) {
+    return true;
+  }
+
+  // Check if application was already running in browser
+  if (from.name) {
+    return false;
+  }
```
simonhir, line 52: Shouldn't this at least show some Snackbar error (error redirect to forbidden page)?

frontend/src/routes/get-started.vue
simonhir, line 1: Shouldn't the menu entry be hidden when the user has none of the roles?

devtobi: You are right, however I just added a simple demo for using `useHasAnyRole` and the route guards, so in my opinion not really necessary here.
We could however move the `useHasAnyRole` example to the `TheNavigationDrawer` component to fulfill your suggestion. However to not have two places for demonstrating the composable, I'd then remove the added dummy text in `index.vue`.

frontend/src/plugins/router.ts
devtobi: We decided to not do that in order to not provide to much information to the user. That's why the redirect to start page only happens on application launch and otherwise no navigation is done when having insufficient roles.

frontend/src/routes/get-started.vue
simonhir: I would do both as in real applications. Otherwise it's only half an example.

frontend/src/plugins/router.ts
simonhir: Why? This leads to bad UX and i think there is nothing bad about showing at least an notification. Security by obscurity is an anti-pattern.

langehm: This is not security by obscurity (that was never the intension) ...
If programmed correctly, you will never be able to navigate inside the application to a page which you cannot access. The only case would be by directly accessing it via url e.g. you got a link to some page.
This would be the only case where a message maybe could make some sense.

frontend/src/plugins/router.ts
```diff
@@ -22,12 +26,32 @@ const router = createRouter({
...
+    hasAnyRole(to.meta.hasAnyRole, userInfoStore.currentRoles)
+  ) {
+    return true;
+  }
+
+  snackbarStore.push({
+    color: STATUS_INDICATORS.ERROR,
+    text: "Du hast nicht die erforderlichen Berechtigungen, um diese Seite anzuzeigen.",
+    icon: mdiShieldLock,
+  });
```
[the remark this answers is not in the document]
devtobi: Not relevant here, will be handled in https://github.com/it-at-m/refarch-templates/issues/1529

simonhir (approved):
LGTM

